Issue #2, opened 2015-07-23
Allow InputStream or Reader as Parser input


robinst:
Currently, the only input type is `String`. In case the input is read from a stream, this means it has to be read into a `String` first, and then the parser has another copy of the data as block content.

Accepting an `InputStream` (or `Reader`?) instead would allow to get rid of one of the copies. `DocumentParser` already processes the input line by line, so this should be trivial.

armstnp:
What would be the viability of passing in more focused type, e.g. `Iterable<String>` or a custom interface such as a `LineSupplier`?

(And yes, the viability of `Iterable` for this type of abstraction is debatable, too.)

robinst:
That would make the common case of parsing a file or other stream of characters harder, because the caller would have to adapt a `BufferedReader` to an `Iterable<String>`. Also, the contract of an `Iterable` is that it can be iterated multiple times, which is not the case for a stream.

After some more thinking, I think we should add an overloaded `parse` with a `Reader` argument. If it's already a `BufferedReader` (e.g. from `Files.newBufferedReader`), we use it as-is, otherwise we wrap it in one. Then read line by line. The caller is responsible for closing it.

I think a `Reader` is better suited than `InputStream`, as we don't want to know about the `Charset`, the caller is responsible for that (also matches [spec](http://spec.commonmark.org/0.21/#character)).

robinst:
There may also be situations where `Iterable<String>` makes sense, but we can add that later if needed.

robinst:
Thanks for working on this, @sentyaev. It would have been nice if you commented here first though, maybe @armstnp would have liked to implement this.

armstnp:
All's good. :+1:

robinst:
Implemented with a new method `parseReader` on `Parser`, thanks @sentyaev

robinst:
Released in [0.2.0](https://github.com/atlassian/commonmark-java/releases/tag/commonmark-parent-0.2.0).

sentyaev:
Great! My first open source project commit.

robinst:
Oh, congratulations! Hopefully there will be many more :).

PR #5, opened 2015-07-27
#2 Reader as input for Parser

sentyaev:
I've add the ability to pass Reader to Parser.

Sorry for 'throws IOException', I'm a .NET developer, 
just don't know how you (Java developers) handle this.

Please help me handle this IOException in Java way, and I'll clean it up and will add tests.

Please do not be rude, I'm a .NET developer, but just starting with Java)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,22 @@ Extensions
 ----------
 
 Extensions need to extend the parser, or the HTML renderer, or both. To
-configure an extension, the builder objects have a method for passing a list of
-extensions. Let's look at how to enable tables (from GitHub Flavored Markdown):
+use an extension, the builder objects can be configured with a list of
+extensions. Because extensions are optional, they live in separate
+artifacts, so additional dependencies need to be added as well.
+
+Let's look at how to enable tables from GitHub Flavored Markdown.
+First, add an additional dependency (see [Maven Central] for others):
+
+```xml
+<dependency>
+    <groupId>com.atlassian.commonmark</groupId>
+    <artifactId>commonmark-ext-gfm-tables</artifactId>
+    <version>0.1.0</version>
+</dependency>
+```
+
+Then, configure the extension on the builders:
 
 ```java
 import org.commonmark.ext.gfm.tables.TablesExtension;
@@ -74,9 +88,8 @@ HtmlRenderer renderer = HtmlRenderer.builder().extensions(extensions).build();
 
 To configure another extension in the above example, just add it to the list.
 
-The following extensions are developed with this library. Note that they are not
-in the same artifact as the core library, so you will need to add additional
-dependencies.
+The following extensions are developed with this library, each in their
+own artifact.
 
 ### Autolink
 


commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
```diff
@@ -1,14 +1,14 @@
...
 import org.commonmark.internal.util.Parsing;
-import org.commonmark.internal.util.Substring;
 import org.commonmark.node.*;
 import org.commonmark.parser.block.*;
 
 import java.util.*;
 
-public class DocumentParser implements ParserState {
-
+public abstract class DocumentParser<T> implements ParserState {
```
robinst: I think adding a class hierarchy for this is a bit too complicated. I'd prefer to just have another method on `DocumentParser` that takes a `Reader`. The code before `incorporateLine` can be moved into the constructor, the part after into a single method that needs to be called by both `parse` methods.

commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
```diff
@@ -63,31 +63,19 @@ public DocumentParser(List<BlockParserFactory> blockParserFactories, InlineParse
...
-
+        
+        incorporateLines(input);
+        
         finalizeBlocks(activeBlockParsers, lineNumber);
         this.processInlines();
         return documentBlockParser.getBlock();
     }
-
+
```
robinst: What editor are you using? It shouldn't add unnecessary spaces on empty lines.

commonmark/src/main/java/org/commonmark/internal/ReaderParser.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class ReaderParser extends DocumentParser<Reader> {
+
+    public ReaderParser(List<BlockParserFactory> blockParserFactories, InlineParserImpl inlineParser) {
+        super(blockParserFactories, inlineParser);
+    }
+    
+    @Override
+    protected void incorporateLines(Reader input) throws IOException {
+        try (BufferedReader r = new BufferedReader(input)){
```
robinst: This causes the passed `input` to be closed at the end of the block, which is not what we want. We want the caller to handle this, we will just read from it until we're done. The typical caller then looks like this:

```java
try (BufferedReader reader = Files.newBufferedReader(filePath)) {
    Node node = parser.parse(reader);
    /* ... */
}
```

Also, I think we should do the following:

```java
BufferedReader bufferedReader;
if (input instanceof BufferedReader) {
    bufferedReader = (BufferedReader) input;
} else {
    bufferedReader = new BufferedReader(input);
}
```

commonmark/src/main/java/org/commonmark/internal/ReaderParser.java
```diff
@@ -0,0 +1,24 @@
...
+            String line;
+            while((line = r.readLine()) != null) {
```
robinst: Missing space after `while`.

commonmark/src/main/java/org/commonmark/parser/Parser.java
```diff
@@ -39,16 +43,26 @@ public static Builder builder() {
      * @param input the text to parse
      * @return the root node
      */
-    public Node parse(String input) {
+    public Node parse(String input) throws IOException {
```
robinst: The `throws IOException` is not really necessary, see my comments above.

commonmark/src/test/java/org/commonmark/test/UsageExampleTest.java
```diff
@@ -10,11 +12,20 @@
...
         Parser parser = Parser.builder().build();
         Node document = parser.parse("This is *Sparta*");
         HtmlRenderer renderer = HtmlRenderer.builder().escapeHtml(true).build();
         assertEquals("<p>This is <em>Sparta</em></p>\n", renderer.render(document));
     }
+    
+    @Test
+    public void two() throws IOException {
+        Parser parser = Parser.builder().build();
+        StringReader sr = new StringReader("This is *Sparta*");
```
robinst: Let's add a test that uses a real file, e.g. `spec.txt`. You can do `UsageExampleTest.class..getResourceAsStream("/spec.txt")` to get an InputStream for the file. Compare it to rendering the same file using a `String` as an input.

Also, this class is for the examples that are used in `README`. Could you move it to a new `ParserTest` instead?

robinst:
Thanks for the pull request :). See my inline comments.

Also, you will have to sign the following Contributor License Agreement (CLA) before I can accept the pull request (I'm sorry, but this is currently required):

https://na2.docusign.net/Member/PowerFormSigning.aspx?PowerFormId=3f94fbdc-2fbe-46ac-b14c-5d152700ae5d

commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
sentyaev: It's not my editor, it's me. I think I do not check for empty line, or just miss it.

commonmark/src/test/java/org/commonmark/test/UsageExampleTest.java
sentyaev: sure

sentyaev:
I'm sign form you provide.

Thanks for reply.

Your comments are reasonable.
So updates will come soon.

PR #6, opened 2015-07-28
#2 Reader to input for Parser

sentyaev:
So, for now it's clean and simple)
Just another method and no inheritance.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
@@ -1,5 +1,8 @@
 package org.commonmark.internal;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.Reader;
 import org.commonmark.internal.util.Parsing;
 import org.commonmark.internal.util.Substring;
 import org.commonmark.node.*;
@@ -75,6 +78,27 @@ public class DocumentParser implements ParserState {
         }
         if (input.length() > 0 && (lineStart == 0 || lineStart < input.length())) {
             incorporateLine(Substring.of(input, lineStart, input.length()));
+        }
+
+        finalizeBlocks(activeBlockParsers);
+        this.processInlines();
+        return documentBlockParser.getBlock();
+    }
+    
+    public Document parse(Reader input) throws IOException {
+        DocumentBlockParser documentBlockParser = new DocumentBlockParser();
+        activateBlockParser(documentBlockParser);
+
+        BufferedReader bufferedReader;
+        if (input instanceof BufferedReader) {
+            bufferedReader = (BufferedReader) input;
+        } else {
+            bufferedReader = new BufferedReader(input);
+        }
+        
+        String line;
+        while ((line = bufferedReader.readLine()) != null) {
+            incorporateLine(line);
         }
 
         finalizeBlocks(activeBlockParsers);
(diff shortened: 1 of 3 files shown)


commonmark/src/main/java/org/commonmark/parser/Parser.java
```diff
@@ -48,6 +50,16 @@ public Node parse(String input) {
         }
         return document;
     }
+    
+    public Node parse(Reader input) throws IOException {
```
robinst: I'm now thinking maybe we shouldn't use overloading here. So maybe rename it to `parseReader` instead, so that it's clearly separate.

commonmark/src/main/java/org/commonmark/parser/Parser.java
```diff
@@ -48,6 +50,16 @@ public Node parse(String input) {
...
+        InlineParserImpl inlineParser = new InlineParserImpl(specialCharacters, delimiterCharacters, delimiterProcessors);
+        DocumentParser documentParser = new DocumentParser(blockParserFactories, inlineParser);
+        Node document = documentParser.parse(input);
+        for (PostProcessor postProcessor : postProcessors) {
+            document = postProcessor.process(document);
+        }
+        return document;
```
robinst: Could you extract the above 4 lines into a private method and also use it from the other `parse` method? Otherwise we're duplicating this.

commonmark/src/test/java/org/commonmark/test/ParserTest.java
```diff
@@ -0,0 +1,37 @@
...
+            int ch;
+            while ((ch = reader.read()) != -1){
+                sb.append((char)ch);
+            }
+        }
+        
+        Node document2 = parser.parse(sb.toString());
+        
+        HtmlRenderer renderer = HtmlRenderer.builder().escapeHtml(true).build();
+        assertEquals(renderer.render(document2), renderer.render(document1));
```
robinst: Good test!

commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
```diff
@@ -81,6 +84,27 @@ public Document parse(String input) {
         this.processInlines();
         return documentBlockParser.getBlock();
     }
+    
+    public Document parse(Reader input) throws IOException {
+        DocumentBlockParser documentBlockParser = new DocumentBlockParser();
+        activateBlockParser(documentBlockParser);
```
robinst: If we put those two lines into the contructor, then we don't have to repeat them in both parse methods.

commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
```diff
@@ -81,6 +84,27 @@ public Document parse(String input) {
...
+        }
+        
+        String line;
+        while ((line = bufferedReader.readLine()) != null) {
+            incorporateLine(line);
+        }
+
+        finalizeBlocks(activeBlockParsers);
+        this.processInlines();
+        return documentBlockParser.getBlock();
```
robinst: These three lines could also be shared by extracting them into a method and reusing it.

robinst:
Looking better! I've added some minor comments, the rest looks good now. Note that you can just add additional commits and then push them, no need to create a new pull request for these changes.

(By the way, sorry for being unresponsive, I'm on holidays right now.)

commonmark/src/main/java/org/commonmark/parser/Parser.java
sentyaev: Hm... From my point of view it's ths same logic.
Please share your thoughts why you think or prefer make it separate.

sentyaev:
No problem! Thanks for your review.
I'll take care of this in couple days.

And have a good holidays!

sentyaev:
So, I've updete code. Check it out.

robinst:
Thanks! :+1:

Issue #3, opened 2015-07-23
Allow rendering to OutputStream/Writer


robinst:
Similar to #2, rendering to a stream should be possible. Also fairly simple to implement.

robinst:
Actually, there is already a method that accepts an `Appendable` (which `Writer` implements), so this is already implemented.

PR #16, opened 2015-09-03
#3 to OutputStream

sentyaev:
Added render to OutputStream

### DIFF
--- a/commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java
@@ -1,5 +1,9 @@
 package org.commonmark.html;
 
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import org.commonmark.Extension;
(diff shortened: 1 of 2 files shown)


commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java
```diff
@@ -46,6 +50,12 @@ public void render(Node node, Appendable output) {
         RendererVisitor rendererVisitor = new RendererVisitor(new HtmlWriter(output), customHtmlRenderers);
         node.accept(rendererVisitor);
     }
+    
+    public void render(Node node, OutputStream output) throws IOException {
+        Writer writer = new OutputStreamWriter(output);
```
robinst: This uses the default character encoding, which is almost never what you actually want (except on Linux, where it's UTF-8).

On second thought, I think the existing method with `Appendable` is enough; `Writer` also implements that, so the functionality is already there. Sorry that this has caused you to do unnecessary work :(. I'll close #3.

sentyaev: Thats fine.

PR #20, opened 2015-09-22
Support for <AutoLink> and link reference definition associations.

pcj:
- An abstract node 'Reference' that captures the link label, title,
  description attributes for subclasses Link, Image, AutoLink.
- AutoLink node class that allows Visitor to differentiate between
  inline links and <autolink> types.
- Support in the InlineParserImpl to capture the association between
  links and link reference definitions.
- static Debugging functions

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/util/Debugging.java
+++ b/commonmark/src/main/java/org/commonmark/internal/util/Debugging.java
@@ -1,4 +1,4 @@
-package org.commonmark.util;
+package org.commonmark.internal.util;
 
 import java.util.Stack;
 import org.commonmark.node.Node;
--- a/commonmark/src/main/java/org/commonmark/internal/util/Debugging.java
+++ b/commonmark/src/main/java/org/commonmark/internal/util/Debugging.java
@@ -1,0 +1,58 @@
+package org.commonmark.util;
+
+import java.util.Stack;
+import org.commonmark.node.Node;
+
+public class Debugging {
+
+    /**
+     * Print a tree-representation of the given node and its children.
+     */
+    public static String toStringTree(Node node) {
+        StringBuilder b = new StringBuilder();
+        Stack<Node> stack = new Stack();
+
+        visit(b, stack, node);
+
+        return b.toString();
+    }
+
+    private static void visit(StringBuilder b, Stack<Node> stack, Node node) {
+
+        int sz = stack.size();
+        for (int i = 0; i < sz; i++) {
+            b.append('.');
+        }
+        b.append(node.toString());
+        b.append('\n');
+
+        Node current = node.getFirstChild();
+        while (current != null) {
+            stack.push(current);
+            visit(b, stack, stack.peek());
+            stack.pop();
+            current = current.getNext();
+        }
+
+    }
+
+    /**
+     * Log a simple message.  This is not a substitute for a logging
+     * framework.
+     */
+    public static void log(String msg) {
+        System.out.println("commonmark-java: " + msg);
+    }
+
+    /**
+     * Print a stacktrace with the given message.
+     */
+    public static void stacktrace(String msg) {
+	try {
+	    throw new Exception(msg);
+	} catch (Exception ex) {
+	    ex.printStackTrace();
+	}
+    }
+
+}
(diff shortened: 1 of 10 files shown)


pcj:
Looking into that test failure.

commonmark/src/main/java/org/commonmark/node/AutoLink.java
```diff
@@ -0,0 +1,17 @@
+package org.commonmark.node;
+
+public class AutoLink extends Link {
```
robinst: It makes sense that `AutoLink` is a separate node type.

I'm not sure about subclassing `Link` though. After all, it doesn't have all the attributes that a `Link` has (no label or title). Have you thought about having it extend `Node` instead?

commonmark/src/test/java/org/commonmark/test/LinkReferenceDefinitionTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class LinkReferenceDefinitionTest {
+
+    @Test
+    public void one() {
+
+        final Parser.Builder builder = Parser.builder();
+        final Parser parser = builder.build();
+        final Node document = parser.parse(getText());
+        final java.util.List<Link> links = new java.util.ArrayList();
```
robinst: Add imports for `java.util.List` and `java.util.ArrayList`. Also, please add generics (`new ArrayList<>();`).

commonmark/src/test/java/org/commonmark/test/LinkReferenceDefinitionTest.java
```diff
@@ -0,0 +1,64 @@
...
+        Link one = links.get(0);
+        Link two = links.get(1);
+        Link auto = links.get(2);
+
+        assertEquals("1", one.getDefinition().getLabel());
+        assertEquals("2", two.getDefinition().getLabel());
+        assertEquals(AutoLink.class, auto.getClass());
+   }
+
+    String getText() {
```
robinst: Make this private please.

commonmark/src/test/java/org/commonmark/test/LinkReferenceDefinitionTest.java
```diff
@@ -0,0 +1,64 @@
...
+        String s = "";
+        s += "* this is collapsed link: [one][1]\n";
+        s += "* this is shortcut link: [2]\n";
+        s += "* this is an AutoLink: <http://foo.com> ...\n";
+        s += "* not to be confused with the *Autolink plugin*,";
+        s += "  that is not picked up in this test: http://foo.com\n";
+        s += "\n";
+        s += "[1]: http://1.com\n";
+        s += "[2]: http://2.com\n";
+        return s;
```
robinst: It doesn't really matter in this case, but can you change this to a single statement instead?:

```java
return "* this is collapsed link: ..." +
        "...";
```

commonmark/src/test/java/org/commonmark/test/LinkReferenceDefinitionTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        assertEquals(3, links.size());
+
+        Link one = links.get(0);
+        Link two = links.get(1);
+        Link auto = links.get(2);
+
+        assertEquals("1", one.getDefinition().getLabel());
+        assertEquals("2", two.getDefinition().getLabel());
+        assertEquals(AutoLink.class, auto.getClass());
```
robinst: Nice test!

commonmark/src/main/java/org/commonmark/internal/util/Debugging.java
```diff
@@ -0,0 +1,58 @@
...
+        b.append('\n');
+
+        Node current = node.getFirstChild();
+        while (current != null) {
+            stack.push(current);
+            visit(b, stack, stack.peek());
+            stack.pop();
+            current = current.getNext();
+        }
+
```
robinst: Let's not leave empty lines at the beginning and end of methods.

robinst:
Looks like a good change, but I'm not sure about the class structure yet.

Do you think it would make sense to have `[foo]: /url "title"` be its own Node type, like `ReferenceDefinition`? Something like this maybe:

* `[foo](/uri "title")` is a `Link`, which has `getDestination`, `getTitle` (not sure about `getLabel`)
* `[foo][foo]`, `[foo][]` and `[foo]` are `ReferenceLink extends Link`, which have an additional `getDefinition` which returns the `ReferenceDefinition`

Does that make sense?

robinst:
I've created an issue for this improvement, for which this PR can be a starting point: #98. Closing this in the meantime.

PR #21, opened 2015-09-22
Adds support for introspection on the delimiter character and length

pcj:
Adds support for introspection on the delimiter character and length for
inline elements emphasis (italic), strong emphasis (bold), and
strikethrough nodes.  New interface 'Delimited' can be implemented by
other custom node classes with similar behaviour.


pcj:
Strange fails only on java8.  @robinst can you check this?  I'm still on java7 and would rather not change at the moment.

robinst:
Looks like the branch also includes the changes for #20. I'll just review the single commit for now.

robinst:
@pcj I've separated the commit for delimiter introspection from the others and implemented my suggestion on top, see pull request #36. Closing this one.

PR #24, opened 2015-10-09
Metadata Extension

chiwanpark:
This pull request contains following changes:

* Add a maven submodule called `commonmark-ext-metadata`
* Add an extension which enables metadata feature
* Add metadata documentation into `README.md`

The metadata syntax is inspired by YAML format. But this extension supports only few features of YAML because of avoiding other YAML dependency and implementing easily.

This is a first pull request which I send to atlassian open source. So there may be many problems to merge this. Please feel free to leave comments. ;)


README.md
```diff
@@ -142,6 +142,26 @@ Enables tables using pipes as in [GitHub Flavored Markdown][gfm-tables].
 
 Use class `TablesExtension` in artifact `commonmark-ext-gfm-tables`.
 
+### Metadata
+
+Enables metadata block in YAML format. This extension only supports limited features. The example is following:
```
robinst: Replace "This extension only supports limited features" with "This extension only supports a subset of YAML syntax"? And replace "The example is following:" with "Here's an example of what's supported:" to make it clear that this is it.

commonmark-ext-metadata/pom.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>commonmark-parent</artifactId>
+        <groupId>com.atlassian.commonmark</groupId>
+        <version>0.2.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
```
robinst: Can you move this to before the `<parent>`? Not important, but let's keep it consistent between the extensions.

README.md
robinst: Also, is there some kind of specification for this header that we can link to?

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/MetadataVisitor.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MetadataVisitor extends AbstractVisitor {
+    private Map<String, List<String>> data;
+
+    public MetadataVisitor() {
+        data = new HashMap<>();
```
robinst: How about using a `LinkedHashMap` instead? I'm not sure if YAML generally preserves the order of key-value pairs or not, but it seems like it would be appropriate here.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class MetadataBlockParser extends AbstractBlockParser {
+    private static final Pattern REGEX_METADATA = Pattern.compile("^[ ]{0,3}([A-Za-z0-9_-]+):\\s*(.*)");
+    private static final Pattern REGEX_METADATA_LIST = Pattern.compile("^[ ]+-\\s*(.*)");
+    private static final Pattern REGEX_METADATA_LITERAL = Pattern.compile("^\\s*(.*)");
+    private static final Pattern REGEX_BOUNDARY = Pattern.compile("^(-{3,}|\\.{3,})(\\s*)?");
+
+    private List<String> lines;
+    private MetadataBlock block;
+    private boolean literal;
```
robinst: Why is this a field? Seems like it's ony used in `parseInlines`.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,112 @@
...
+                    node.setValues(values);
+                    block.appendChild(node);
+                }
+
+                literal = false;
+                key = matcher.group(1);
+                values = new ArrayList<>();
+                if ("|".equals(matcher.group(2))) {
+                    literal = true;
+                } else if (!"".equals(matcher.group(2))) {
```
robinst: Is this legal?:

```
key: 
```

In that case, the list of values should be empty? Could you add a test case for that?

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,112 @@
...
+                    values.add(matcher.group(2));
+                }
+            } else {
+                matcher = REGEX_METADATA_LIST.matcher(line);
+                if (matcher.matches()) {
+                    values.add(matcher.group(1));
+                } else if (literal) {
```
robinst: Can a literal value start with `-` like this?:

```
---
key: |
  - test
---
```

Can you add a test case for this as well? :)

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,112 @@
...
+            node.setKey(key);
+            node.setValues(values);
+            block.appendChild(node);
+        }
+    }
+
+    public static class Factory extends AbstractBlockParserFactory {
+        @Override
+        public BlockStart tryStart(ParserState state, MatchedBlockParser matchedBlockParser) {
+            String line = state.getLine().toString();
```
robinst: Is the `toString` necessary here? `Pattern#matcher` takes a `CharSequence`.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/MetadataExtension.java
```diff
@@ -0,0 +1,26 @@
+package org.commonmark.ext.metadata;
+
+import org.commonmark.Extension;
+import org.commonmark.ext.metadata.internal.MetadataBlockParser;
+import org.commonmark.ext.metadata.internal.MetadataBlockRenderer;
+import org.commonmark.html.HtmlRenderer;
+import org.commonmark.parser.Parser;
+
+public class MetadataExtension implements Parser.ParserExtension, HtmlRenderer.HtmlRendererExtension {
```
robinst: Could you add some Javadoc here? Link to the node types and the visitor.

pom.xml
```diff
@@ -26,6 +26,7 @@
         <module>commonmark-ext-gfm-tables</module>
         <module>commonmark-integration-test</module>
         <module>commonmark-test-util</module>
+        <module>commonmark-ext-metadata</module>
```
robinst: Move this up next to the other extensions please.

robinst:
Hey! Very good work, I only have some minor comments, see above. Some more comments:

* [ ] Can metadata appear in the middle of the document, or does it have to be at the beginning? In case it has to be at the beginning, maybe we should check for that in `tryStart`?
* [ ] Please add the extension to `SpecIntegrationTest`, so that we know we don't break anything in core with this extension.
* [ ] Do you feel that this level of support for YAML covers almost all- use cases of document in the wild?

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
chiwanpark: Is `parseInlines` called once? If right, we can move `literal` into `parseInlines`.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
chiwanpark: Yes, this is legal case. I added a test case to cover this.

chiwanpark:
Sorry for late update. I updated my PR and rebased on current master branch. About questions:

* Metadata cannot appear in the middle of the document. I already added a test case covering that (`metadataInParagraph`).
* I added metadata extension to `SpecIntegrationTest`. The integration test passed.
* I'm not sure that this extension covers almost all use cases. But this extension provides features which are similar to GFM and python markdown library (https://pythonhosted.org/Markdown/extensions/meta_data.html).

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
robinst: It is. Yes, please move it :).

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,147 @@
...
+
+                if (validLineCount > 0) {
+                    return BlockStart.of(new MetadataBlockParser()).atIndex(state.getNextNonSpaceIndex());
+                }
+            }
+
+            return BlockStart.none();
+        }
+
+        private int nextLineEnd(CharSequence seq, int startIndex) {
```
robinst: This won't work. `tryStart` only gets one line of the document. I think checking that we're at the beginning of the document and that we have a boundary marker is enough.

Also, please add a test for the following input:

```
---
test
```

That currently fails because there's an infinite loop in `tryStart`.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,147 @@
...
+            if (matcher.matches()) {
+                continue;
+            }
+
+            matcher = REGEX_METADATA.matcher(line);
+            if (matcher.matches()) {
+                if (key != null) {
+                    MetadataNode node = new MetadataNode();
+                    node.setKey(key);
+                    node.setValues(values);
```
robinst: I'd add a constructor that takes key and values (in addition to the default constructor), because the same code is repeated at the end of this method.

chiwanpark:
I addressed your comments @robinst. :)

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
robinst: I had another look, and the fact that this is working is actually a bug in the `Substring` class, it doesn't do the necessary bounds checking. I'm going to fix these bugs on master, which will mean that the above will stop working. But it wouldn't work when reading the input from a file anyway (only one line at a time).

You'll need to merge master and adjust the implementation. Thanks for your patience :).

robinst: Ok, pushed to master. Could you merge master into your branch and adapt?

chiwanpark: If we check only one line to decide whether metadata block starts or not, there are some collisions with CommonMark spec like following case:

```
---
test
----
```

From original CommonMark spec, rendering result should be `<hr />\n<h2>test</h2>\n`. But because this extension consider `---` as start of metadata block, the actual result will be empty. Is there any method to use multi-line input for detecting start of the block?

robinst: There isn't currently. If the first line is a normal paragraph line, then there's support for replacing it on the second line with a block. That is used e.g. for headlines, where only the second line (with the `-`) shows that it's a headline. But in the above case, the first line is parsed as a horizontal rule right away, not a paragraph.

Do you know how e.g. the Python implementation handles this?

Are there more collisions? If we can't find a way, we can exclude the examples in `SpecIntegrationTest`. I'm not sure it's a problem in practice, because it only affects the very first line. And if a user encounters this case, they can work around it by adding an empty line before it or an empty metadata block using two `...` lines.

chiwanpark: From my investigation, the Python implementation (https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/meta.py) receives all unparsed-markdown string and check whole lines to decide whether metadata block starts or not.

If we check block boundary using only a line, example 14 and 16 in Setext headers of `SpecIntegrationTest` fail.

chiwanpark: By the way, the Python implementation and github markdown implementation allow only `---` in first line to begin a metadata block. I would like to apply this rule in this extension.

robinst: Thanks for investigating!

Looks like the Python library implements it as a pre-processor that removes the metadata block from the input. Then later the rest of the document is parsed. I'm a bit hesitant to explore that direction because it makes source positions more complicated. Also, that doesn't really work nicely when parsing from a Reader instead of a String.

So let's go with keeping it simple for now, and if the first line in the document happens to be a `---`, we interpret it as a metadata block. We could do some more checking in `tryContinue` and stop the block if the first line of content doesn't look like metadata content.

That means we'll need to override example 14 and 16 in `SpecIntegrationTest`, see `getOverriddenExamples`.

Also, +1 for only allowing `---`.

chiwanpark:
Hi @robinst, sorry for late update. I added a block verification logic to `tryContinue` method and overridden examples.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class MetadataBlockParser extends AbstractBlockParser {
+    private static final Pattern REGEX_METADATA = Pattern.compile("^[ ]{0,3}([A-Za-z0-9_-]+):\\s*(.*)");
+    private static final Pattern REGEX_METADATA_LIST = Pattern.compile("^[ ]+-\\s*(.*)");
+    private static final Pattern REGEX_METADATA_LITERAL = Pattern.compile("^\\s*(.*)");
+    private static final Pattern REGEX_BEGIN = Pattern.compile("^-{3}(\\s.*)?");
+    private static final Pattern REGEX_END = Pattern.compile("^(-{3}|\\.{3})(\\s.*)?");
```
robinst: Is this intentional, allowing `...` to end the block but not to start it?

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Override
+    public BlockContinue tryContinue(ParserState parserState) {
+        final CharSequence line = parserState.getLine();
+
+        if (REGEX_END.matcher(line).matches()) {
+            // if this line is `---` or `...` which means end of metadata block
+            return BlockContinue.finished();
+        } else if (!REGEX_METADATA.matcher(line).matches() && !REGEX_METADATA_LIST.matcher(line).matches() &&
+                !REGEX_METADATA_LITERAL.matcher(line).matches()) {
```
robinst: This regex is `^\\s*(.*)`, won't this just always match? Neither space nor any character is required.

What I meant with doing the matching is moving the logic from `parseInlines` to `tryContinue`. Instead of iterating over the lines in `parseInlines`, keep the state in the class and process one line at a time. If a line doesn't match anything, finish the block.

I can take a stab at it if you want.

chiwanpark: @robinst Oh I misunderstood your comments. I'll modify the implementation.

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
robinst: Replying to myself here, but looks like other implementations also do that: https://github.com/jgm/lcmark/#yaml-metadata

So disregard my comment :).

commonmark-ext-metadata/src/main/java/org/commonmark/ext/metadata/internal/MetadataBlockParser.java
robinst: :+1:

derari:
Just a quick side note: Wouldn't calling this extension "YAML front matter" (because that's what it is, even if not all YAML is supported) be more descriptive than just calling it "Metadata"? There is probably more than one way one might encounter metadata when handling Markdown, so I think it might be helpful for user if it was clear from the extension's name where and what kind of metadata it supports.

robinst:
@derari: I think that's a good point. @chiwanpark, when you come around to updating this PR, could you make that change? Other than that, we're really close, and I'd love to see this included in the next release :).

chiwanpark:
I'm really sorry for delaying update @robinst. I'll update this PR including @derari's suggestion in 3 days.

About a name of maven module, is `commonmark-ext-yaml-frontmatter` reasonable? I found that a similar implementation in Jekyll is called `frontmatter` also (http://jekyllrb.com/docs/frontmatter/).

robinst:
Cool! I'd add another hyphen, `commonmark-ext-yaml-front-matter`, because it's multiple words.

chiwanpark:
Hi all, I updated this PR addressing your opinions. Please review this. :-)

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/yaml/YAMLFrontMatterBlock.java
```diff
@@ -0,0 +1,6 @@
+package org.commonmark.ext.yaml;
```
robinst: Can you change the packages to `org.commonmark.ext.front.matter` instead? (Seems more future proof, apparently there's other front matter formats, e.g. JSON.)

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/yaml/YAMLFrontMatterBlock.java
```diff
@@ -0,0 +1,6 @@
...
+
+import org.commonmark.node.CustomBlock;
+
+public class YAMLFrontMatterBlock extends CustomBlock {
```
robinst: Can you change the `YAML` in all class names to `Yaml`? This is to make it consistent with `Html` in `HtmlRenderer`.

chiwanpark:
Thanks for review @robinst. I just addressed your comments.

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YAMLFrontMatterBlock.java
```diff
@@ -0,0 +1,6 @@
+package org.commonmark.ext.front.matter;
+
+import org.commonmark.node.CustomBlock;
+
+public class YamlFrontMatterBlock extends CustomBlock {
```
robinst: Looks like the file names aren't changed, this file is still named `YAMLFrontMatterBlock`. That's why the Travis build fails. Are you on a case-insensitive file system? Check out this [Stackoverflow question](http://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git) for how to work around that.

chiwanpark:
Hi @robinst, I updated the all filenames. Thanks for your guide.

robinst:
Finally merged :smile:. Nice work Chiwan!

chiwanpark:
Thanks @robinst and @derari!

robinst:
Released in 0.5.0 now :tada:

PR #33, opened 2016-01-25
Avoid ClassNotFoundException on Android (API level < 19)

JinneeJ:
There was removed usage of following classes: `Objects` and `StandardCharset`. When support of Android below KitKat (API level 19) becomes [not-actual](http://developer.android.com/intl/ru/about/dashboards/index.html) these changes can be reverted.


robinst:
Thanks :).

PR #34, opened 2016-01-26
Update license year to range

pra85:


robinst:
Thanks for the patch. Could do the same for the year in README.

PR #38, opened 2016-02-08
Test android project

JinneeJ:
This PR contains basic Android project to test compatibility  of `commonmark-java` on Android with SDK version >= 15.

Also was added README for commonmark-android-test with instructions and links :)

Currently `autolinkExtensionTest` fails for SNAPSHOT because of `Objects`. There are 2 options: deleting test of autolink or adding android support to autolink :)
All tests for current maven version fail - as expected.

Also was fixed SpecReader since android tests use the spec.


robinst:
Wow, this looks great :smiley:! I'll try to look into how we can use Travis or some other CI solution to include this in the build, hopefully soon.

For autolink, does this fix the problem?: https://github.com/robinst/autolink-java/commit/a8e85e36421e99f0a179066a209160c2beeaf5b4

JinneeJ:
It looks that this change will fix the problem :)

robinst:
Released a new version of autolink-java and bumped it on master. Can you check if it works?

Also started trying to make the build on Travis work on the [test-android-project-travis](https://github.com/atlassian/commonmark-java/commits/test-android-project-travis) branch. It currently fails. Here's the result I get, any ideas what the problem could be?:

https://travis-ci.org/atlassian/commonmark-java/jobs/108276173

> commonmark-android-test/app/build.gradle' line: 44

I'll try turning on some more debugging later.

JinneeJ:
**Autolink**
With new autolink-java all tests are passed 👍

**Travis**
It seems that you don't create `test.properties`. May be it will work after adding to `.travis.yml` something like that:

```shell
...cd commonmark-android-test
&& touch ./test.properties
&& echo "path.report=../report" > ./test.properties
&& echo "version.maven=0.4.0" >> ./test.properties
&& echo "version.snapshot=0.4.1-SNAPSHOT" >> ./test.properties
&& echo "version.snapshot_autolink=0.4.0" >> ./test.properties
&& ./gradlew :app:connectedSnapshotDebugAndroidTest; fi
```

Also if Travis doesn't support the launch of Android emulator then you need to add the launch of emulator to `.travis.yml` too (please see [README](https://github.com/Doist/commonmark-android/blob/test-android-project/commonmark-android-test/README.md))

robinst:
Merged. I've done the Travis changes in a follow-up PR, see #40.

PR #41, opened 2016-02-18
Fix bullets with tabs

derari:
A fix for this similar bug from commonmark.js: https://github.com/jgm/commonmark.js/issues/86


robinst:
The fix looks good, thanks. Before I can merge it, could you please fill out the following form (we currently require it for first-time contributors):

https://na2.docusign.net/Member/PowerFormSigning.aspx?PowerFormId=3f94fbdc-2fbe-46ac-b14c-5d152700ae5d

robinst:
Thanks for the pull request :). (Sorry that it took some time, should be much quicker for subsequent ones.)

Issue #26, opened 2015-10-28
An option to add automatic IDs to headings


Wilfred:
GitHub flavoured markdown adds automatic IDs to headings, and it would be great to see this as an option here (disclaimer: I'm hoping to see the feature in Stash/Bitbucket server).

This is not in commonmark ([and won't be added](http://talk.commonmark.org/t/feature-request-automatically-generated-ids-for-headers/115)) though [some implementations even do this by default](http://johnmacfarlane.net/babelmark2/?text=%23+Heading).

In the case of GitHub, a heading `#foo` gets an ID `user-generated-foo` to avoid clobbering IDs used elsewhere. There's then some JS that means that you can use `#foo` in your URL instead of `#user-generated-foo`.

This would allow us to use tools like [markdown-toc](https://github.com/jonschlinkert/markdown-toc) or [doctoc](https://github.com/thlorenz/doctoc) on Bitbucket server.

robinst:
Cheeky :). But I agree, this would be useful.

Should be pretty simple to build it as an extension using an [`AttributeProvider`](https://github.com/atlassian/commonmark-java/blob/master/commonmark/src/main/java/org/commonmark/html/AttributeProvider.java) that looks at `Header` nodes.

I won't have time in the next few weeks, but if someone else wants to take a stab at this, you're welcome.

robinst:
I've looked into what existing implementations do for converting the heading text to an ID:

http://johnmacfarlane.net/babelmark2/?text=%23+Heading%0A%0A%23+heading%0A%0A%23+heading%252%0A%0A%23+heading{}3%0A%0A%23+heading+%C3%A4%0A%0A%23+heading+%F0%9F%98%84%0A%0A%23+heading+%F0%9F%98%84%0A%0A%23+heading+%F0%9F%98%84%0A%0A%23+heading++4%0A%0A%23+1+heading%0A%0A%23+123%0A%0A%23+--%0A%0A%23+00A1%C2%A1test%0A%0A%23+A2%C2%A2test

HTML itself doesn't put any restrictions on the format, but other specifications do. One that is probably useful to make IDs conform to is CSS:

http://www.w3.org/TR/CSS21/syndata.html#value-def-identifier

Looks like that is what Pandoc does, except that it strips more characters such as ¢ (U+00A2).

It also avoids duplicates by appending `-N`.

vsch:
[...]

PR #42, opened 2016-02-19
Add header ids to all h tags

paulthom12345:
This is a hesitant pull request. I want to do more testing around international characters and ensuring that there aren't some weird edge cases. Still thought it's better to show progress early to get any feedback on things that might not be to your vision.


paulthom12345:
I've got to clean this up (code formatting) and fix merge conflicts.

robinst:
Finally got to review this, sorry that it took so long. So, this is a good start! Comments here:

1. We should aim for having the same behavior as GitHub in general. See vsch's comment in #26 for some hints:

  * The implementation should not only handle ASCII, but check the Unicode properties. One way to do it would be using `Character.isAlphabetic`. Or use a `Pattern` to match all the allowed characters. `bär` should result in the ID `bär`. While we're at it, it should also be able to handle `a` followed by U+0308 COMBINING DIAERESIS.

  * Test that `# heading\n# heading\n# heading-1` results in the IDs `heading`, `heading-1`, `heading-1` (I think that's already the case due to how the code works).

2. I think it would be useful if the logic of calculating the IDs (including keeping a map of previous IDs) was in its own self-contained public class. That way, users can reuse it even if they want to do something else than add an `ID` attribute to the tag (some implementations add an inner `<a>` element for example).

3. Given that the change adds a new artifact for this with `-ext` in its name, it should probably also provide an `FooExtension` class that registers the `AttributeProvider`.

paulthom12345:
I think that should have fixed up what you've outlined. Let me know what other improvements should be made. I didn't go as far as to support every unicode combining character. If you'd like to do that, do you know of any general approach that's a good idea, or should I just be using unicode ranges like i've done so far?

paulthom12345:
The checks seem to be on Licenses for the internal modules. Would you like me to set a license on the module i've added?

robinst:
1. Please move the code from `IdAttribute` to `UniqueIdentifierProvider` as well, that's the most tricky part of the logic. Just collect the Text and Code literals into one string (separated by spaces) and then pass it do `UniqueIdentifierProvider` and let it do all the replacement.

2. Getting all the right character classes is tricky. I would use the `UNICODE_CHARACTER_CLASS` flag of `Pattern` instead:

   ```java
            Pattern p = Pattern.compile("[\\w-]+", Pattern.UNICODE_CHARACTER_CLASS);
            Matcher matcher = p.matcher(s);
            while (matcher.find()) {
                sb.append(matcher.group());
            }
   ```

   There's some more test cases that work with this:

        assertRendering("# a\u1DC0", "<h1 id=\"a\u1DC0\">a\u1DC0</h1>\n");
        assertRendering("# undertie \u203F", "<h1 id=\"undertie-\u203F\">undertie \u203F</h1>\n");

3. It looks like GitHub converts `foo  bar` to `foo--bar` (doesn't collapse spaces). We should do the same.

4. Can you change the artifact name to `ext-heading-anchor` and the package to `ext.heading.anchor`? I looked at extensions for other libraries and they all had "anchor" in the name.

Don't worry about the build, that's something they (versioneye) introduced just recently without asking. I've removed the hooks that do this now, should go away next time you push.

paulthom12345:
The only thing (i think) that i haven't quite done is the `ext-heading-anchor` artifact id. Each of the other artifact ids are prefixed with `commonmark-` so therefore i set it as `commonmark-ext-heading-anchor`. If you'd rather I switched from convention i'm happy to do so.

README.md
```diff
@@ -143,6 +143,18 @@ Enables tables using pipes as in [GitHub Flavored Markdown][gfm-tables].
...
+
+Enables adding auto generated id attributes to header based on their content.
+
+`# Heading` will be rendered as
+
+```
+<h1 id="heading">Heading</h1>
+```
+
+Use class `HeaderIdExtension` in artifact `commonmark-ext-header-ids`
```
robinst: s/commonmark-ext-header-ids/commonmark-ext-heading-anchor/

commonmark-ext-heading-anchor/src/main/java/ext/heading/anchor/HeaderIdExtension.java
```diff
@@ -0,0 +1,45 @@
+package ext.heading.anchor;
```
robinst: Sorry, I wasn't clear. I meant only to change the part after the `org.commonmark`. The package should be `org.commonmark.ext.heading.anchor`.

commonmark-ext-heading-anchor/src/main/java/ext/heading/anchor/HeaderIdExtension.java
```diff
@@ -0,0 +1,45 @@
...
+ * </code>
+ * will result in
+ * <code>
+ * <pre>
+ * &lt;h1 id="heading"&gt;Heading&lt;/h1&gt;
+ * &lt;h1 id="heading-1"&gt;Heading&lt;/h1&gt;
+ * </pre>
+ * </code>
+ */
+public class HeaderIdExtension implements HtmlRenderer.HtmlRendererExtension {
```
robinst: Can you rename this to `HeadingAnchorExtension` as well?

commonmark-ext-heading-anchor/src/main/java/ext/heading/anchor/internal/HeaderIdAttributeProvider.java
```diff
@@ -0,0 +1,51 @@
...
+import org.commonmark.html.AttributeProvider;
+import org.commonmark.node.AbstractVisitor;
+import org.commonmark.node.Code;
+import org.commonmark.node.Heading;
+import org.commonmark.node.Node;
+import org.commonmark.node.Text;
+
+import ext.heading.anchor.UniqueIdentifierProvider;
+
+public class HeaderIdAttributeProvider implements AttributeProvider {
```
robinst: This one should be named `HeadingIdAttributeProvider` (the nodes are called `Heading` after the naming was changed in the spec).

commonmark-ext-heading-anchor/src/main/javadoc/overview.html
```diff
@@ -0,0 +1,6 @@
+<html>
+<body>
+<b>Extension for automatically adding Id attributes to all headers</b>
+<p>See {@link org.commonmark.ext.autolink.AutolinkExtension}</p>
```
robinst: This link is wrong.

paulthom12345:
Only just realised i had to do the upstream sync! Sorry!

PR #44, opened 2016-03-11
Corrected wrong package name in javadoc overview

jiakuan:
Sample change


robinst:
Nice catch, thanks :+1:

Issue #35, opened 2016-02-01
Override rendering functions


md384:
Hi. Is there a way to override the rendering functions? For example, I'd like to render a link differently if it links to page on our domain vs one off-domain.

I was thinking to override the `RenderVisitor` inner class in `HtmlRenderer` but it has `private` access. Is there another way to do this?

csauve:
We had a similar use case of needing to modify the rendering of certain tags. For example we wanted to avoid SEO impact of the user-provided texts on pages, so links get `rel="nofollow"` and headers get rendered as styled paragraphs.

I found that you can't override the rendering of built-in node types, and need to replace them with nodes extending `CustomNode` or `CustomBlock` for your `CustomHtmlRenderer` to get called. I implemented a visitor extending `AbstractVisitor` and overrode `visit` for the tags I wanted to render differently. For example (groovy):
```groovy
    @Override
    void visit(Link link) {
      def replacement = new CustomLink()
      replacement.destination = link.destination
      replacement.title = link.title
      replacement.appendChild(link.firstChild)
      link.insertAfter(replacement)
      link.unlink()
    }
```

`CustomLink` is a `CustomNode` that can be rendered specially:
```groovy
    @Override
    boolean render(Node node, HtmlWriter htmlWriter, Visitor visitor) {
      if (node instanceof CustomLink) {
        htmlWriter.tag("a", [
            "href": node.destination,
            "target": "_blank",
            "rel": "nofollow"
        ])
        visitChildren(node, visitor)
        htmlWriter.tag("/a")
        return true
      }
      return false
    }

    private static void visitChildren(Node node, Visitor visitor) {
      Node child = node.getFirstChild()
      while (child != null) {
        child.accept(visitor)
        child = child.getNext()
      }
    }
```

So the steps are:
1. Parse the markdown as normal
2. Run a visitor on the node tree that replaces the nodes you want rendered differently with custom node types
3. Implement a `CustomHtmlRenderer` for the render phase that outputs your custom nodes the way you want

Hope this helps!

md384:
@csauve Thanks, that works great!

csauve:
Agreed, it would be nice if the `CustomHtmlRenderer` implementations could be used for any arbitrary node type. The order in which they're added determines the precedence. The default built-in renderer would only kick in if no custom renderer returned `true` for handling the type.

@robinst, is this a misuse of the API and does there exist a simpler way to do this?

robinst:
[...]

Issue #31, opened 2016-01-25
AttributeProvider not working for TablesExtension and images


rmuller:
As documented in the source code (`TableHtmlRenderer#renderBlock()`):

````
// TODO: What about attributes? If we got the renderer instead of the visitor, we could call getAttributes.
````

To be able to set custom attributes would be very handy especially when working with Tables!

More in general: IMO `AttributeProvider` should be working for all tags, including images.

robinst:
Thanks for the report! Yeah, let's fix it properly and also make it work for images.

In the case of images, we currently don't support it because we use the same renderer to render the alt attribute (see [here](https://github.com/atlassian/commonmark-java/blob/master/commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java#L310)). I'm thinking we should change that and create a new renderer, then take the rendered string for the alt value, put it in the attributes and then let the attribute provider extend it.

robinst:
Are you interested in contributing this yourself? Otherwise I'll pick it up when I get to it :).

robinst:
Fixed with #49.

robinst:
Released in 0.5.0.

PR #49, opened 2016-04-08
Make HTML rendering for nodes extensible (#35)

robinst:
We had `CustomHtmlRenderer` for custom nodes before. That was ok for
that, but didn't allow clients to override rendering for core nodes.

This introduces a new concept `NodeRenderer` (including factories, who
doesn't like those?) that is also used for the core nodes. That allows
overriding the rendering and allows to define the rendering for custom
nodes at the same time.

NodeRenderers have access to rendering configuration and functionality,
such as extending attributes. That fixes the problem that the renderer
for table nodes wasn't extensible via `AttributeProvider` (#31).


commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java
```diff
@@ -148,8 +150,18 @@ public Builder attributeProvider(AttributeProvider attributeProvider) {
...
+         * Add a factory for instantiating a node renderer (done when rendering). This allows to override the rendering
+         * of node types or define rendering for custom node types.
+         * <p>
+         * If multiple node renderers for the same node type are created, the one from the factory that was added first
+         * "wins". (This is how the rendering for core node types can be overridden; the default rendering comes last.)
+         *
+         * @param nodeRendererFactory the factory for creating a node renderer
+         * @return {@code this}
+         */
+        public Builder nodeRendererFactory(NodeRendererFactory nodeRendererFactory) {
```
robinst: This is the new extension point. See also [NodeRenderer](https://github.com/atlassian/commonmark-java/pull/49/files#diff-f6dfcbbe0c786ca98aae0f5a8c8e6498) and associated interfaces.

commonmark-ext-gfm-strikethrough/src/main/java/org/commonmark/ext/gfm/strikethrough/internal/StrikethroughNodeRenderer.java
```diff
@@ -0,0 +1,44 @@
...
+import org.commonmark.html.HtmlWriter;
+import org.commonmark.html.renderer.NodeRenderer;
+import org.commonmark.html.renderer.NodeRendererContext;
+import org.commonmark.node.Node;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+public class StrikethroughNodeRenderer implements NodeRenderer {
```
csauve: Why does the `CoreNodeRenderer` extend `AbstractVisitor` but the extension renderers do not? `renderChildren()` here is reimplmenting `visitChildren()`.

commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java
csauve: Looks great! Overriding is simple and intuitive.

commonmark-ext-gfm-strikethrough/src/main/java/org/commonmark/ext/gfm/strikethrough/internal/StrikethroughNodeRenderer.java
robinst: Because this one is about the `Strikethrough` node, which is not a built-in node. That's why it doesn't work with visitor.

Other than that, yeah, `renderChildren` is kind of re-implemented. Maybe `NodeRendererContext` should have a `renderChildren` method. But for now I'd rather keep it minimal.

commonmark/src/main/java/org/commonmark/html/HtmlRenderer.java
robinst: Cool, thanks for reviewing!

koesie10:
Could a new version be released with this change included?

koesie10:
Great, thanks!

PR #58, opened 2016-08-03
Add text content renderer

JinneeJ:
Finally added ability to get clean text.

P.S.: Really sorry for big delay 🙏


robinst:
Nice change!

Do you have any plans for making it work for custom nodes for extensions (e.g. `Strikethrough`)? Maybe we should have a similar model as for HTML rendering where extensions can contribute code for rendering, and users can customize the rendering for specific node types. See `NodeRenderer` and friends.

(By the way, I was on holidays, that's why it took so long to respond.)

commonmark/src/main/java/org/commonmark/content/TextContentRenderer.java
```diff
@@ -0,0 +1,253 @@
...
+                    textContent.write("- ");
+                }
+                visitChildren(listItem);
+                writeEndOfLine(listItem, null);
+            }
+        }
+
+        @Override
+        public void visit(OrderedList orderedList) {
+            orderedListCounter = 1;
```
JinneeJ: You're right :) Fixed here https://github.com/atlassian/commonmark-java/pull/58/commits/2d520cc8736c0b0f078f0ffa529dabd479c4ef66

JinneeJ:
Thanks for review! I've made the fix with lists.

Also I've tried to implement support extensions for TextContentRenderer. But I had to make some changes in core in order to avoid copying of code. You can see it in last 5 commits. Let me know what you think about those changes 😊

robinst:
Cool, starting to look good!

I wanted to make a few suggestions but wasn't sure how they'd work, so I checked out the branch and started modifying it :). I merged master into your branch and did these changes on top:

1. https://github.com/atlassian/commonmark-java/commit/43c7e28a7cd32d5fb5a4fb367f4bf7dd7371d627
2. https://github.com/atlassian/commonmark-java/commit/91827e2364f9f3bb162df22cf012d04f6c0980fb

Can you have a look and tell me if it makes sense?

JinneeJ:
I think your changes are great 😊

robinst:
Thanks! Merged now. Will do a release soon with these and other changes.

robinst:
Released now, thanks a lot @JinneeJ: https://github.com/atlassian/commonmark-java/releases/tag/commonmark-parent-0.7.0

Issue #54, opened 2016-07-11
Add underline support


rjoly:
Hello,

Could we add a ```UnderlineExtension``` in a new artifact ```commonmark-ext-gfm-underline```? It would be a cousin class of the existing ```StrikethroughExtension``` in artifact ```commonmark-ext-gfm-strikethrough```, with few little changes on this new class and others related to.

If you agree this issue, I could submit a pull request. However, I just need to know which delimiter should be the more accurate to do so.
I thing we should use an "easy to use" character on both QWERTY and AZERTY keyboards (and some others?), with a double occurrence. Here are some proposals: ```--``` or ```&&``` or ```%%```.

Thank you in advance for your feedback.

robinst:
Hey :).

Can you point me to an existing Markdown implementation that has support for this? From [this StackExchange question](http://programmers.stackexchange.com/questions/207727/why-there-is-no-markdown-for-underline), it looks like this is not really something that is intended.

I'm also interested in your use case, what do you need this for? And is the HTML support not suitable to do this (e.g. use `<u>foo</u>`)?

rjoly:
I had been reading the same StackExchange question and I understand the answer, only regarding the Markdown philosophy (more or less "meaning versus presentation"). I also notice the underline is not easy to read and should introduce some confusion with the conventional rendering of hyperlinks.

In my use case, I want to allow my final user to easily format a text with understandable code (not HTML code). So I think Markdown can help me to do this easily and efficiently. But the underline feature is one of the favourite features needed by these end users (with the 2 existing levels of emphasis which are traduced by italic and bold presentation) and is not in the Markdown perimeter.

I can't use ```<u>foo</u>``` because my end user will forget the ```/``` to close the tag, or make any other mistake. It would also force me to break the user input escaping of my UI framework to rebuild it with my small arms, just for a ```<u>``` tag.

So, as your implementation offers a strikethrough extension, which doesn't seem to be part of the Markdown specification, I suggest to had an underline extension. As I understand, as an extension, it won't change the default implementation which respects the Markdown specification.
I know the ```<del>``` tag is a semantic one contrary to the ```<u>``` one, but I think it should be a useful feature in some case.

Note: looking for markdown implementation with underline, I found a source talking about the ```<ins>``` tag. This little-used tag mean that its content has been inserted and the default presentation is to underline the text.

robinst:
[...]

PR #60, opened 2016-09-05
create "underline <ins>" extension (#54)

pabranch:
copied from "strikethrough &lt;del&gt;" and renamed

Comments and suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,13 @@ Enables tables using pipes as in [GitHub Flavored Markdown][gfm-tables].
 
 Use class `TablesExtension` in artifact `commonmark-ext-gfm-tables`.
 
+### Ins
+
+Enables underlining of text by enclosing it in `++`. For example, in
+`hey ++you++`, `you` will be rendered as underline text. Uses the &lt;ins&gt; tag.
+
+Use class `InsExtension` in artifact `commonmark-ext-ins`.
+
 ### YAML front matter
 
 Adds support for metadata through a YAML front matter block. This extension only supports a subset of YAML syntax. Here's an example of what's supported:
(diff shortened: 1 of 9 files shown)


robinst:
Hey there, welcome! This looks good already, I only have one comment:

I think we should call the artifact, package and class names `ins` instead. Whether it's rendered as underline or something else depends on the CSS that's used in the end, and is out of control of this library. A user of the library might not even render to HTML but just parse text with it. Note that the [MDN entry about ins](https://developer.mozilla.org/en/docs/Web/HTML/Element/ins) says nothing about `ins` being underline. We can keep the mentioning of underline in the descriptions. (I know this contradicts a bit with what the `gfm-strikethrough` extension does, but there it's because we follow exactly what GitHub does and calls it.)

We also require that you sign this CLA (only one time because it's your first contribution): https://na2.docusign.net/Member/PowerFormSigning.aspx?PowerFormId=3f94fbdc-2fbe-46ac-b14c-5d152700ae5d

Once you've done the above change and sign the CLA, I'll gladly merge this :).

pabranch:
I am happy to make the name change. When I have a chance, I'll look over the CLA and send a new pull request. Thanks for such a prompt response!

pabranch:
How does this copy sound for the README.md?

> ### Ins
>
>Enables underline of text by enclosing it in `++`. For example, in
>`hey ++you++`, `you` will be rendered as underline text. Uses the &lt;ins&gt; tag.
>
>Use class `InsExtension` in artifact `commonmark-ext-ins`.

robinst:
Sounds good! Don't forget to change the package name as well :).

pabranch:
CLA has been signed. Let me know if you need me to email you a copy.

robinst:
Perfect, thank you! Will be included in the next release.

PR #66, opened 2016-10-04
Add prefix/suffix to ids generated

paulthom12345:
* Adds the ability for IdGenerator to be passed in prefixes and suffixes
that will ensure to be added in-front/after every id. This will
allow users to 'namespace' their generated ids to a certain pool.

### DIFF
--- a/commonmark-ext-heading-anchor/src/main/java/org/commonmark/ext/heading/anchor/HeadingAnchorExtension.java
+++ b/commonmark-ext-heading-anchor/src/main/java/org/commonmark/ext/heading/anchor/HeadingAnchorExtension.java
@@ -29,11 +29,26 @@ import org.commonmark.renderer.html.HtmlRenderer;
  */
 public class HeadingAnchorExtension implements HtmlRenderer.HtmlRendererExtension {
 
-    private HeadingAnchorExtension() {
+    private final String defaultId;
+    private final String idPrefix;
+    private final String idSuffix;
+
+    private HeadingAnchorExtension(String defaultId, String idPrefix, String idSuffix) {
+        this.defaultId = defaultId;
+        this.idPrefix = idPrefix;
+        this.idSuffix = idSuffix;
     }
 
     public static Extension create() {
-        return new HeadingAnchorExtension();
+        return create(builder());
+    }
+
+    public static Extension create(Builder builder) {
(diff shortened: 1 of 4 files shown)


commonmark-ext-heading-anchor/src/main/java/org/commonmark/ext/heading/anchor/internal/HeadingIdAttributeProvider.java
```diff
@@ -12,12 +12,20 @@
...
-        idGenerator = IdGenerator.builder().defaultId("heading").build();
+    private HeadingIdAttributeProvider(String defaultId, String prefix, String suffix) {
+        idGenerator = IdGenerator.builder()
+                .defaultId(defaultId)
+                .prefix(prefix)
+                .suffix(suffix)
+                .build();
     }
 
     public static HeadingIdAttributeProvider create() {
```
paulthom12345: I'm pretty sure i should remove this `create` method. This isn't part of the public API, and will be unused if using the public API

commonmark-ext-heading-anchor/src/main/java/org/commonmark/ext/heading/anchor/HeadingAnchorExtension.java
```diff
@@ -29,20 +29,78 @@
...
+        this.idPrefix = idPrefix;
+        this.idSuffix = idSuffix;
     }
 
     public static Extension create() {
-        return new HeadingAnchorExtension();
+        return create(builder());
+    }
+
+    public static Extension create(Builder builder) {
```
robinst: This does not need to be public. Instead, just pass the builder to the private constructor. See what we do in `HtmlRenderer`.

commonmark-ext-heading-anchor/src/main/java/org/commonmark/ext/heading/anchor/internal/HeadingIdAttributeProvider.java
robinst: Yes, feel free to remove it 👍.

robinst:
Released with 0.7.1: https://github.com/atlassian/commonmark-java/releases/tag/commonmark-parent-0.7.1

PR #67, opened 2016-10-24
Added ability to include/use only specific core block factories

lalunamel:
via the Parser builder

Signed-off-by: Cody Sehl csehl@pivotal.io

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
@@ -8,18 +8,26 @@ import org.commonmark.internal.util.Substring;
 import org.commonmark.node.*;
 import org.commonmark.parser.block.*;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.LinkedHashSet;
+import java.util.HashSet;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 
 public class DocumentParser implements ParserState {
 
-    private static final List<Class<? extends Block>> CORE_FACTORY_TYPES = Arrays.<Class<? extends Block>>asList(
+    private static final Set<Class<? extends Block>> CORE_FACTORY_TYPES = new LinkedHashSet<>(Arrays.asList(
             BlockQuote.class,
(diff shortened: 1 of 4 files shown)


commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
```diff
@@ -59,11 +73,17 @@ public DocumentParser(List<BlockParserFactory> blockParserFactories, InlineParse
...
 
-    public static List<BlockParserFactory> calculateBlockParserFactories(List<BlockParserFactory> customBlockParserFactories) {
+    public static List<Class<? extends Block>> getDefaultBlockParserTypes() {
+        return CORE_FACTORY_TYPES;
+    }
+
+    public static List<BlockParserFactory> calculateBlockParserFactories(List<BlockParserFactory> customBlockParserFactories, List<Class<? extends Block>> allowedBlockTypes) {
         List<BlockParserFactory> list = new ArrayList<>();
         // By having the custom factories come first, extensions are able to change behavior of core syntax.
         list.addAll(customBlockParserFactories);
```
robinst: I was a bit hesitant of this change because it didn't handle custom block parsers coming from extensions. But then again, the library user is the one actually configuring the extensions, so I don't think this is a problem :).

commonmark/src/main/java/org/commonmark/parser/Parser.java
```diff
@@ -116,6 +118,37 @@ public Builder extensions(Iterable<? extends Extension> extensions) {
...
+         *     Parser.builder().allowedBlockTypes(Heading.class, ListBlock.class);
+         *     }
+         * </pre>
+         *
+         * @param allowedBlockTypes A list of nodes the parser will parse.
+         *                     If this list is empty, the parser will not recognize any Commonmark core markdown features.
+         *
+         * @return {@code this}
+         */
+        public Builder allowedBlockTypes(List<Class<? extends Block>> allowedBlockTypes) {
```
robinst: Wondering if it should be a `Set` instead of a `List`, seems to fit the use better.

robinst:
I've seen that you signed the CLA, excellent! When you address the minor comments I've added before, this will be good to merge.

robinst:
Thanks, merged. Will be included in the next release!

PR #69, opened 2016-10-24
Make inline parsing customizable by providing an implemetaion of Inli…

lalunamel:
…neParser

Signed-off-by: Graham Bell gbell@pivotal.io
Signed-off-by: Cody Sehl csehl@pivotal.io

### DIFF
--- a/commonmark/src/main/java/org/commonmark/parser/InlineParser.java
+++ b/commonmark/src/main/java/org/commonmark/parser/InlineParser.java
@@ -4,7 +4,6 @@ import org.commonmark.node.Node;
 
 /**
  * Parser for inline content (text, links, emphasized text, etc).
- * <p><em>This interface is not intended to be implemented by clients.</em></p>
  */
 public interface InlineParser {
 
@@ -13,5 +12,10 @@ public interface InlineParser {
      * @param node the node to append resulting nodes to (as children)
      */
     void parse(String input, Node node);
+
+    /**
+     * @return how many characters were parsed as a reference, {@code 0} if none
+     */
+    int parseReference(String s);
 
 }
--- a/commonmark/src/main/java/org/commonmark/parser/InlineParser.java
+++ b/commonmark/src/main/java/org/commonmark/parser/InlineParser.java
@@ -12,10 +12,4 @@ public interface InlineParser {
      * @param node the node to append resulting nodes to (as children)
      */
     void parse(String input, Node node);
-
-    /**
-     * @return how many characters were parsed as a reference, {@code 0} if none
-     */
-    int parseReference(String s);
-
 }
--- a/commonmark/src/main/java/org/commonmark/parser/Parser.java
+++ b/commonmark/src/main/java/org/commonmark/parser/Parser.java
@@ -81,7 +81,8 @@ public class Parser {
         if (this.inlineParserFactory == null) {
             return new InlineParserImpl(delimiterProcessors);
         } else {
-            return this.inlineParserFactory.create(delimiterProcessors);
+            CustomInlineParserContext inlineParserContext = new CustomInlineParserContext(delimiterProcessors);
+            return this.inlineParserFactory.create(inlineParserContext);
         }
     }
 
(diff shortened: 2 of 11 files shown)


lalunamel:
Related to https://github.com/atlassian/commonmark-java/issues/68

commonmark/src/main/java/org/commonmark/parser/InlineParser.java
```diff
@@ -14,4 +13,9 @@
      */
     void parse(String input, Node node);
 
+    /**
+     * @return how many characters were parsed as a reference, {@code 0} if none
+     */
+    int parseReference(String s);
```
robinst: I'm not sure if this belongs here or not, because it forces implementations of `InlineParser` to keep the reference map as state and it bakes this ugly part of the parser into the API.

For another issue, we'd also like to preserve parsed link reference definitions as nodes in the AST so that they can be used to render the original Markdown source back. So we'd have to break this API if we wanted to implement that.

Maybe I can look into that soon, but in the meantime, is there a way for you to not add this to the API while still enabling you to customize inline parsing?

lalunamel: We've looked at the intended behavior behind parseReference and agree that it shouldn't be part of the InlineParser's API.
But it looks like the ParagraphParser uses this method when calling `closeBlock`.

We've searched through the code and we're not sure how to best divorce `parseReference` from the InlineParser interface when it's relied upon by ParagraphParser, given that the ParagraphParser is required to update the internal state/referenceMap of the InlineParser.

Any ideas here?

lalunamel:
Hey @robinst, any thoughts here?

lalunamel:
Ping!
(If we get this merged we don't need to manage a local, non-maven version of commonmark on all our dev machines :) )

commonmark/src/main/java/org/commonmark/parser/InlineParser.java
robinst: We could put this method on a separate interface that lives in `.internal` for now, and `ParagraphParser` does an `if (inlineParser instanceof ReferenceParser)` before calling it. That way, we wouldn't have to put it in the API while your extension would still work I think. Thoughts?

That way this PR can go in while I think some more about what a proper `parseReference` should look like.

commonmark/src/main/java/org/commonmark/parser/Parser.java
```diff
@@ -75,12 +77,20 @@ public Node parse(String input) {
...
         DocumentParser documentParser = new DocumentParser(blockParserFactories, inlineParser);
         Node document = documentParser.parse(input);
         return postProcess(document);
     }
 
+    private InlineParser getInlineParser() {
+        if (this.inlineParserFactory == null) {
+            return new InlineParserImpl(specialCharacters, delimiterCharacters, delimiterProcessors);
+        } else {
+            return this.inlineParserFactory.create();
```
robinst: Does your custom inline parser not need to know about `delimiterProcessors` etc? I think the `create` method should get a `InlineParserContext` (similar to `HtmlNodeRendererContext`), that looks like this:

```java
interface InlineParserContext {
    Map<Character, DelimiterProcessor> getDelimiterProcessors();
}
```

(The other two, `specialCharacters` and `delimiterCharacters` can be derived from the delimiter processors and are just an optimization, so don't need to be on the interface.)

lalunamel:
We split out `parseReference` into a separate `ReferenceParser` interface.
While not explicitly required, implementations of InlineParser require implementing both `InlineParser` and `ReferenceParser`, else the ParagraphParser won't work as expected.

We also changed the way custom `delimiterProcessor`s are specified for an InlineParser.
We didn't wrap them in a Context because we felt it was unnecessary.

robinst:
Thanks for making the changes, merged now. Will release with these changes soon!

robinst:
I just released 0.9.0 which includes these changes: https://github.com/atlassian/commonmark-java/releases/tag/commonmark-parent-0.9.0

(Was waiting for another PR to include in the release, but it's not ready yet.)

PR #70, opened 2016-10-30
Fix for Heading without space after # 

codesForLiving:
`#Heading `

Is not parsed as heading now. Space or tab is needed after #

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/HeadingParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/HeadingParser.java
@@ -10,7 +10,7 @@ import java.util.regex.Pattern;
 
 public class HeadingParser extends AbstractBlockParser {
 
-    private static Pattern ATX_HEADING = Pattern.compile("^#{1,6}(?:[ \t]+|$)");
+    private static Pattern ATX_HEADING = Pattern.compile("^#{1,6}(?:[ \t]*|$)");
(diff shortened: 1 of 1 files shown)


codesForLiving:
- [ ] 

- [ ]

robinst:
Looked at the code anyway :). So just a heads-up: The reason it requires a space is so that it conforms to the [commonmark spec](http://spec.commonmark.org/0.26/#atx-headings):

> The opening sequence of # characters must be followed by a space or by the end of line.

So I wouldn't merge a pull request that makes the space optional in all cases. Maybe it could be an option on the parser, but I'm also not sure if that's a good thing to do, I don't want to provide too many non-spec-compliant options either.

codesForLiving:
> Note that many implementations currently do not require the space.

From : [commnmark spec page](http://spec.commonmark.org/0.26/#atx-headings):

> but I'm also not sure if that's a good thing to do

It seems like a good enough reason to provide it as option. :)

robinst:
The "many implementations" in that sentence refers to other Markdown implementations as opposed to CommonMark, see other uses of "implementations" in the spec.

Could you explain what your use case is for which you need this?

codesForLiving:
I need it to parse data from Reddit. 
They don't need space after #. 
I can use my fork for my requirement. Thanks :)

PR #92, opened 2017-08-02
Update text content renderer

JinneeJ:
This PR add some fixes/improvements for text content rendering:
1. Added support for sub lists
2. Fixed that whitespaces between text elements are removed in "stripped" mode. For example `**text** and text` had rendered as `textand text`
3. Improved rendering for auto links

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/text/CoreTextContentNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/text/CoreTextContentNodeRenderer.java
@@ -2,6 +2,8 @@ package org.commonmark.renderer.text;
 
 import org.commonmark.node.*;
 import org.commonmark.renderer.NodeRenderer;
+import org.commonmark.renderer.text.holder.BulletListHolder;
+import org.commonmark.renderer.text.holder.OrderedListHolder;
 
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


commonmark/src/main/java/org/commonmark/renderer/text/CoreTextContentNodeRenderer.java
```diff
@@ -68,15 +68,22 @@ public void visit(BlockQuote blockQuote) {
...
         textContent.write('»');
 
-        writeEndOfLine(blockQuote, null);
+        writeEndOfLineIfNeeded(blockQuote, null);
     }
 
     @Override
     public void visit(BulletList bulletList) {
-        bulletListMarker = bulletList.getBulletMarker();
+        if (bulletListHolder != null) {
```
robinst: What about a bullet list that is nested within a ordered list (or the other way around)? Is that handled correctly?

robinst: @JinneeJ Tested now with the following input:

```
source = "* a\n  1. one\n  2. two\n* b"
```

Renders as this with the default renderer:

```
* a1. one
2. two
* b
```

Can you fix that?

JinneeJ: I overlooked it somehow, sorry! 😅 
Updated PR with according tests and fixes.

robinst:
Thanks! The build failed for an unrelated reason, see fix in #100.

Issue #96, opened 2017-08-25
NPE caused by InlineParsers


spearce:
I'm getting an NPE while parsing a simple document like:

```
foo {} is a "test" right?
```

The NPE seems to be caused by having two InlineParser extensions fire on the same paragraph of text. In this case I have an InlineParser extension on { and ". Code for the extensions is below. The NPE seems to be caused by the InlineParserImpl incorrectly creating the Node its passing into the second extension that fires on the paragraph. The node is not attached to any other node, it has no parent. Its not a member of the paragraph.

```
java.lang.NullPointerException
	at org.commonmark.node.Node.insertAfter(Node.java:89)
	at com.google.gitiles.doc.SmartQuotedExtension.quote(SmartQuotedExtension.java:51)
	at com.google.gitiles.doc.SmartQuotedExtension.access$0(SmartQuotedExtension.java:43)
	at com.google.gitiles.doc.SmartQuotedExtension$QuotedProcessor.process(SmartQuotedExtension.java:86)
	at org.commonmark.internal.InlineParserImpl.processDelimiters(InlineParserImpl.java:854)
	at org.commonmark.internal.InlineParserImpl.parse(InlineParserImpl.java:169)
	at org.commonmark.internal.ParagraphParser.parseInlines(ParagraphParser.java:61)
	at org.commonmark.internal.DocumentParser.processInlines(DocumentParser.java:410)
	at org.commonmark.internal.DocumentParser.finalizeAndProcess(DocumentParser.java:529)
	at org.commonmark.internal.DocumentParser.parse(DocumentParser.java:118)
	at org.commonmark.parser.Parser.parse(Parser.java:60)
```

``` java
public class NamedAnchorExtension implements ParserExtension {
  @Override
  public void extend(Parser.Builder builder) {
    builder.customDelimiterProcessor(new Processor());
  }

  private static class Processor implements DelimiterProcessor {
    @Override
    public char getOpeningCharacter() {
      return '{';
    }

    @Override
    public char getClosingCharacter() {
      return '}';
    }

    @Override
    public int getMinLength() {
      return 1;
    }

    @Override
    public int getDelimiterUse(DelimiterRun opener, DelimiterRun closer) {
      return 1;
    }

    @Override
    public void process(Text opener, Text closer, int delimiterUse) {
      // Silly yes, but simplified for this issue report.
      opener.insertAfter(new Text("{"));
      closer.insertBefore(new Text("}"));
    } 
}

[...]

PR #97, opened 2017-08-28
Fix text node merging when opening/closing delimiters are adjacent (#96)

robinst:
When using asymmetric delimiters, e.g. `{` as an opener and `}` as a
closer, the input `{}` means there are no nodes between the delimiters.
So we have the following nodes:

1. `{` (opener)
2. `}` (closer)

The code to merge text nodes was merging nodes starting from opener.next
and ending with closer.previous. In the above situation, that would mean
merging starting from "2" (and ending at "1"). Because it never got to
node "1", it would just keep merging nodes until the end of the block.

The NPE in #96 is a possible symptom of that. In other cases, delimiter
processing might lead to the wrong result.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
+++ b/commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
@@ -175,7 +175,7 @@ public class InlineParserImpl implements InlineParser, ReferenceParser {
         } while (moreToParse);
 
         processDelimiters(null);
-        mergeTextNodes(block.getFirstChild(), block.getLastChild());
+        mergeChildTextNodes(block);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
```diff
@@ -958,7 +958,25 @@ private void removeDelimiter(Delimiter delim) {
         }
     }
 
-    private void mergeTextNodes(Node fromNode, Node toNode) {
+    private void mergeTextNodesBetweenExclusive(Node fromNode, Node toNode) {
+        // No nodes between them
+        if (fromNode == toNode || fromNode.getNext() == toNode) {
```
robinst: Not merging on `fromNode.getNext() == toNode` here is the fix.

spearce:
This fixes the sample file, and the original I derived the bug report from. Thanks!

robinst:
@spearce thanks for checking, merged.

PR #101, opened 2017-09-06
Move text renderer util classes to internal package

robinst:
They are not API, so they should be in an internal package.

@JinneeJ small post-merge cleanup, please review :).


JinneeJ:
All looks good!

PR #115, opened 2017-12-08
Issue 80: TextContent extension for GFM tables

ahjaworski:
This implements a text content renderer to render GFM tables to plain text. We use this renderer in our e-mail template engine in which we use Markdown with GFM tables and it gets converted to HTML and plain text to send in e-mails. I don't know what the best default renderer output should be, but this can be changed.


commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/HtmlTablesExtension.java
```diff
@@ -20,13 +20,13 @@
  * The parsed tables are turned into {@link TableBlock} blocks.
  * </p>
  */
-public class TablesExtension implements Parser.ParserExtension, HtmlRenderer.HtmlRendererExtension {
+public class HtmlTablesExtension implements Parser.ParserExtension, HtmlRenderer.HtmlRendererExtension {
```
robinst: It should not be necessary to split `TablesExtension` into two classes, just keep it as is and implement `TextContentRendererExtension` here. See `StrikethroughExtension` for an example.

That also means we won't break existing users of `TablesExtension`.

commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/HtmlTableNodeRenderer.java
```diff
@@ -8,12 +8,12 @@
 
 import java.util.*;
 
-public class TableNodeRenderer implements NodeRenderer {
+public class HtmlTableNodeRenderer implements NodeRenderer {
```
robinst: Please name it `TableHtmlNodeRenderer` (keeping it consistent with `StrikethroughHtmlNodeRenderer`)

commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TextContentTableNodeRenderer.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void render(Node node) {
+        // We don't render the table header (node instanceof TableHead) and its children for the text content.
+
+        if (node instanceof TableBlock) {
+            renderBlock((TableBlock) node);
+        } else if (node instanceof TableHead) {
+            renderHead((TableHead) node);
```
robinst: This seems to contradict the comment at the beginning of this method.

commonmark-ext-gfm-tables/src/test/java/org/commonmark/ext/gfm/tables/TextContentTablesTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Test
+    public void separatorCanNotHaveLeadingSpaceThenPipe() {
+        assertRendering("Abc|Def\n |---|---", "Abc|Def\n|---|---");
+    }
+
+    @Test
+    public void headerMustBeOneLine() {
+        assertRendering("No\nAbc|Def\n---|---", "No\nAbc|Def\n---|---");
+    }
```
robinst: This test and the ones above are not necessary, they test parsing instead of rendering. You can remove them from here. This class should only be about the text rendering.

commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TextContentTableNodeRenderer.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private void renderRow(TableRow tableRow) {
+        textContentWriter.line();
+        renderChildren(tableRow);
+        textContentWriter.line();
+    }
+
+    private void renderCell(TableCell tableCell) {
+        renderChildren(tableCell);
+        textContentWriter.colon();
```
robinst: This is a bit of an odd choice to me. If you have a table with three colums, it would be rendered like this:

```
a: b: c
```

Why did you choose ":" as a separator? A pipe seems more natural even in text.

ahjaworski:
I've processed your review comments. The extension is now separated in its own module and the delimiter is changed to a | (pipe symbol). I took the colon from our own implementation in which we mostly used two-column tables for e-mails that were shown with : as delimiter. I agree a | (pipe symbol) is better for common use. Let me know if it is now conforming the project setup.

ahjaworski:
Indeed, I completely misunderstood your first review comment, luckily the logic was easy to move around. I think I've now done it the way you meant in the first place. I've merged the master branch in this branch and I would think it is now ready for the coming release. I also signed the form.

commonmark/src/main/java/org/commonmark/renderer/text/TextContentWriter.java
```diff
@@ -24,6 +24,12 @@ public void colon() {
         }
     }
 
+    public void pipe() {
```
robinst: Looks great now! The only thing unsure about now is this method. I'm not sure about adding it to the public API, it seems very arbitrary. Is there a good reason to avoid writing two pipes in a row? Otherwise I'd just do `append('|')` in `renderCell` and remove this method.

ahjaworski: It will both have the same outcome I guess. I don't know which solution is most in line with the codebase, so you may decide. With the TablesTextContentTest.java testclass you can make sure the output is still the same.

robinst:
Merged and made the changes on top of it. Thanks :)! Will be in the next release: https://github.com/atlassian/commonmark-java/blob/master/CHANGELOG.md#0110---2018-01-16

ahjaworski:
Nice! Thanks for giving feedback.

PR #117, opened 2017-12-15
Make TextContentWriter extensible

ldrozdz:


commonmark/src/main/java/org/commonmark/renderer/text/TextContentWriter.java
```diff
@@ -64,4 +64,12 @@ private void append(char c) {
 
         lastChar = c;
     }
+
+    protected Appendable getBuffer() {
```
robinst: Would a method like `append(String s)` work for you instead?

ldrozdz: Thanks for getting back to me! For my particular use case, I actually needed access to _lastChar_ in order to override the logic of _line()_. I have managed to work around it, so this PR is not urgent personally - jsut thought it could be a useful addition.

robinst: So you only need `lastChar`? I'm fine with just exposing that. If you update your pull request (or create a new one), I'll merge that.

Issue #84, opened 2017-03-15
Check GFM extensions against new spec


robinst:
GitHub just posted on their blog that GitHub-flavored Markdown is now CommonMark + extensions and has a spec: https://githubengineering.com/a-formal-spec-for-github-markdown/

Spec lives here: https://github.github.com/gfm/

Check our implementation of tables, strikethrough (and maybe autolinking) against the spec.

robinst:
Did tables as part of #149. Checked others too:

* strikethrough: Not much in spec, no need to test against that :)
* autolinking: Too different from what we have at the moment

So I'm closing this.

robinst:
Released as part of 0.13.0: https://twitter.com/niborst/status/1150946272976621569

AfzalivE:
> autolinking: Too different from what we have at the moment

Any chance this has gotten easier to implement since 2019?

PR #122, opened 2018-02-21
Update table delimiter row to GFM spec

jleider:
The GFM spec states that any table delimiter row is valid if it contains at
least one `-` and no other characters except for whitespace and `:`.

See: https://github.github.com/gfm/#tables-extension-

I have signed the CLA for individuals. 

This addresses part of #84

### DIFF
--- a/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java
+++ b/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java
@@ -12,7 +12,7 @@ import java.util.regex.Pattern;
 
 public class TableBlockParser extends AbstractBlockParser {
 
-    private static String COL = "\\s*:?-{3,}:?\\s*";
+    private static String COL = "\\s*:?-{1,}:?\\s*";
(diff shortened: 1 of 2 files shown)


robinst:
Looks good, merged! Thank you :).

Have you looked at running the examples from [their spec](https://github.github.com/gfm/#tables-extension-) in the tests yet? We should be able to use the `SpecReader` and only test the examples with that specific section.

jleider:
@robinst I had not looked at running their examples, however with just a quick glance, it looks like most of their examples are tested in one way or another with the current test suite except one: https://github.github.com/gfm/#example-197. One difference I noticed is that the GFM specs are generous with newline usage, each element is output on its own line which differs from this lib although I don't think that difference is material. Lastly, in the example without a `<tbody>`, the spec says there shouldn't be `<tbody>` tags where this lib outputs an empty `<tbody></tbody>` tag set. Its possible that this could cause issues if padding/margin/height/width CSS styles are applied to `tbody`.

robinst:
I still think it would be nice to run the same tests, just in case we don't cover one of the cases. We can strip whitespace between HTML elements when comparing the results to account for the differences in newlines.

The last one is interesting, we should remove the `<tbody></tbody>`. Do you want to create another pull request?

PR #129, opened 2018-06-15
Configured maven bundle plgin to produce osgi bundles

ieugen:
* Fixes https://github.com/atlassian/commonmark-java/issues/128 
* Over the next days I will test the bundles and see if any changes from the default are needed
* Waiting for a release :)

### DIFF
--- a/commonmark/pom.xml
+++ b/commonmark/pom.xml
@@ -8,6 +8,7 @@
     </parent>
 
     <artifactId>commonmark</artifactId>
+    <packaging>bundle</packaging>
     <name>commonmark-java core</name>
     <description>Core of commonmark-java (implementation of CommonMark for parsing markdown and rendering to HTML)</description>
 
@@ -41,6 +42,11 @@
                         </manifestEntries>
                     </archive>
                 </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
             </plugin>
         </plugins>
     </build>
--- a/commonmark/pom.xml
+++ b/commonmark/pom.xml
@@ -47,6 +47,14 @@
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
                 <extensions>true</extensions>
+                <configuration>
+                  <instructions>
+                    <Export-Package>
+                      *,
+                      org.commonmark.internal
+                    </Export-Package>
+                  </instructions>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/commonmark-ext-autolink/pom.xml
+++ b/commonmark-ext-autolink/pom.xml
@@ -8,6 +8,7 @@
     </parent>
 
     <artifactId>commonmark-ext-autolink</artifactId>
+    <packaging>bundle</packaging>
     <name>commonmark-java extension for autolinking</name>
     <description>commonmark-java extension for turning plain URLs and email addresses into links</description>
 
@@ -45,6 +46,11 @@
                         </manifestEntries>
                     </archive>
                 </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
             </plugin>
         </plugins>
     </build>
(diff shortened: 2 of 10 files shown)


ieugen:
After this I will start working on a simple Karaf feature to easily install the bundles in Karaf.

commonmark/pom.xml
```diff
@@ -8,6 +8,7 @@
     </parent>
 
     <artifactId>commonmark</artifactId>
+    <packaging>bundle</packaging>
```
robinst: I'm a bit uncomfortable with changing the packaging type here. Is there a way to add the manifest entries without changing this?

commonmark/pom.xml
```diff
@@ -42,6 +43,19 @@
...
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                  <instructions>
+                    <Export-Package>
+                      *,
+                      org.commonmark.internal
+                    </Export-Package>
```
robinst: What about the other artifacts, don't they need something like this as well? They also have `internal` packages that should not be exported.

commonmark/pom.xml
```diff
@@ -42,6 +43,19 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
```
robinst: What does this do, is it needed?

commonmark/pom.xml
ieugen: Actually there is, but adds more configuration. However there are no issues with bundle packaging. The plugin exists for a lot of time and works well with all other plugins.

[1] https://bnd.bndtools.org/tools/felix-maven.html

commonmark/pom.xml
ieugen: I haven't got to using extensions yet. If those internal packages are not used by other bundles then no need to export them. I think the whiteboard pattern might fit for extensions. Using that pattern extensions can make themselves available to any user. 

[1] https://en.wikipedia.org/wiki/Whiteboard_Pattern

commonmark/pom.xml
ieugen: This plugin updates the MANIFEST.MF file with osgi metadata.

commonmark/pom.xml
robinst: They are not used by other bundles, yeah.

commonmark/pom.xml
robinst: I meant what does the `<extensions>true</extensions>` do?

ieugen: Yes it is needed. Please see docs: https://maven.apache.org/guides/mini/guide-using-extensions.html .

That is how the plugin configuration is done for it to work.

ieugen:
> Does it automatically use "-" for the non-groupId part?
The algorithm for choosing a name is described in the docs. It works very well. Symbolic name is used to identify the bundle by the framework and for humans. If you are not happy with it you can change it. I don't ever recall changing that in all my work. 

Imports and exports usually require some attention. Other than that, defaults work very well 99% of the time for me.

robinst:
Thanks for your patience with this. I'll merge this if:

* You make it work without changing `<packaging>` (I don't want to make it harder to set up for any IDEs, etc)
* You also go through the extensions to add the appropriate `Export-Package` there

> The algorithm for choosing a name is described in the docs. It works very well. Symbolic name is used to identify the bundle by the framework and for humans. If you are not happy with it you can change it. I don't ever recall changing that in all my work.

Ok, let's leave it. Just looked a bit weird because it only got a `-` at the end :).

josephw:
>     * You make it work without changing `<packaging>`

FWIW, the example in  the [Apache Felix Maven Bundle Plugin documentation](https://bnd.bndtools.org/tools/felix-maven.html#real-world-example) shows that this is a relatively small change:

    <plugin>
      <artifactId>maven-jar-plugin</artifactId>
      <configuration>
        <archive>  
          <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
        </archive> 
      </configuration>
    </plugin>  
    <plugin>   
      <groupId>org.apache.felix</groupId>
      <artifactId>maven-bundle-plugin</artifactId>
      <executions>
        <execution>
          <id>bundle-manifest</id>
          <phase>process-classes</phase>
          <goals>    
            <goal>manifest</goal>
          </goals>   
        </execution>
      </executions>
    </plugin>

With this project's module structure, adding that once to `commonmark-parent` may be enough to get OSGi bundles without modifying the packaging type.

robinst:
Closing this. Open to a less invasive PR that does it the way @josephw pointed out.

PR #130, opened 2018-06-17
Added Karaf feature

ieugen:
This builds on previous PR https://github.com/atlassian/commonmark-java/pull/129 . 
I have tested the feature installation. Please review and release. I would like to re-use and build on top. 
My next goal would be to expose services to parse/render markdown.

### DIFF
--- a/commonmark-ext-autolink/pom.xml
+++ b/commonmark-ext-autolink/pom.xml
@@ -12,10 +12,6 @@
     <name>commonmark-java extension for autolinking</name>
     <description>commonmark-java extension for turning plain URLs and email addresses into links</description>
 
-    <properties>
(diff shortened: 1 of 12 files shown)


robinst:
Can you explain what this is useful for?

I'm not sure about including it in commonmark-java. It seems very specific to some system and could probably live in a different repository for people that need it.

ieugen:
It is useful for installing comonmark  easily inside Apache Karaf application server (and I think also JBoss Fuse and some other tools).

robinst:
Hey, thanks for your patience. I've thought about this and decided that I don't want to include this, because:

* I wouldn't know how to maintain it
* I don't think it's a problem for many users. Even for Karaf users, I'm sure they can handle a library without this feature.

PR #132, opened 2018-06-17
Declarative services dependency for karaf feature

ieugen:
* depend on scr feature to bring up declarative services automatically
* this build on previous PR's #130 , #131


robinst:
Similar to https://github.com/atlassian/commonmark-java/pull/131#issuecomment-405460021, I'm declining this.

PR #137, opened 2018-09-09
Optimize block parsing by replacing regexes

robinst:
Block parsers use regex matching to check whether a line matches that block. That means for every line, we try a bunch of regexes. These can add up. The use of `Matcher` showed up in a profiling session on Android.

One way to fix that would be to check only one character first, and only if it matches what the block parser expects, run the full regex. But that seems like a bit of a strange mix.

Instead, this PR replaces all the regex use in block parsers with fairly simple looping code.

This gives a nice speedup for the benchmarks. Before:

```
Benchmark                 Mode  Cnt    Score   Error  Units
SpecBenchmark.wholeSpec  thrpt  200  109.766 ± 0.366  ops/s
SpecBenchmark.examples   thrpt  200  257.182 ± 0.789  ops/s
```

After:

```
Benchmark                 Mode  Cnt    Score   Error  Units
SpecBenchmark.wholeSpec  thrpt  200  127.930 ± 0.600  ops/s
SpecBenchmark.examples   thrpt  200  305.566 ± 0.959  ops/s
```

So that's about a 15% to 18% improvement! 🎉

There's probably some potential for improvement in inline parsing as well, but haven't looked at that yet.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/FencedCodeBlockParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/FencedCodeBlockParser.java
@@ -100,20 +100,16 @@ public class FencedCodeBlockParser extends AbstractBlockParser {
             }
         }
         if (backticks >= 3 && tildes == 0) {
-            for (int i = index + backticks; i < line.length(); i++) {
-                // spec: The info string may not contain any backtick characters.
-                if (line.charAt(i) == '`') {
-                    return null;
-                }
+            // spec: The info string may not contain any backtick characters.
+            if (Parsing.find('`', line, index + backticks) != -1) {
+                return null;
             }
             return new FencedCodeBlockParser('`', backticks, indent);
         } else if (tildes >= 3 && backticks == 0) {
-            for (int i = index + tildes; i < line.length(); i++) {
-                // This follows commonmark.js but the spec is unclear about this:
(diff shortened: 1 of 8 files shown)


robinst:
Here's some numbers from Android. Before:

```
Simple (Parse): avg=0.05ms median=0.04ms min=0.04ms max=0.05ms
Short (Parse): avg=0.28ms median=0.28ms min=0.26ms max=0.32ms
Medium (Parse): avg=2.55ms median=2.53ms min=2.42ms max=2.78ms
Long (Parse): avg=27.90ms median=27.67ms min=26.69ms max=29.35ms
UltraLong (Parse): avg=671.34ms median=667.70ms min=662.73ms max=712.41ms
```

After:

```
Simple (Parse): avg=0.05ms median=0.05ms min=0.04ms max=0.05ms
Short (Parse): avg=0.28ms median=0.27ms min=0.26ms max=0.31ms
Medium (Parse): avg=1.84ms median=1.84ms min=1.75ms max=1.97ms
Long (Parse): avg=20.05ms median=19.84ms min=19.50ms max=20.91ms
UltraLong (Parse): avg=602.48ms median=599.91ms min=589.54ms max=647.74ms
```

For simple and short only have a single paragraph. Medium and long have multiple.

So it looks like with text that has a lot of blocks, this reduces parsing time by about 30%, nice :).

PR #146, opened 2018-12-27
Add a huge horizontal rule testcase

benhumphreys:
In version 0.11.0 a huge horizontal rule of thousands of characters caused a stack overflow error. In 0.12.0 that was fixed in commit ad397cb79 when ThematicBreakParser.isThematicBreak() was modified to
avoid using a regex.

This regression test fails on 0.11.0 and passes on 0.12.0 and later. With a default stack size (default for my system at least) the test starts failing somewhere between 1000 and 2000 characters. This was reported initially against Bitbucket Server, where some string that I can only imagine wasn't intended to be a real markdown horizontal rule was being parsed as such. See: https://jira.atlassian.com/browse/BSERV-11497


robinst:
Thanks Ben! Glad to see that replacing regexes fixes bugs as well as improving performance :).

PR #156, opened 2019-07-12
Update to commonmark spec 0.29

### LINKED ISSUE #152: Update to CommonMark spec 0.29
Release notes: https://github.com/commonmark/commonmark-spec/releases/tag/0.29

Spec: https://spec.commonmark.org/0.29/
Changes: https://spec.commonmark.org/0.29/changes.html

robinst:
This was a larger one because of the change in link reference definitions, see commonmark/commonmark-spec#395. I also cleaned up parsing of definitions to avoid parsing them twice. We can now also add them as nodes to the document, which is nice.

Fixes #152.


commonmark/src/main/java/org/commonmark/internal/ReferenceParser.java
```diff
@@ -1,11 +0,0 @@
-package org.commonmark.internal;
-
-/**
- * Parser for inline references
- */
-public interface ReferenceParser {
```
robinst: FYI @lalunamel this has been removed (to be released). It shouldn't really impact you, but just a heads up.

PR #159, opened 2019-11-06
Add ability to render into something other than string

OlegKrikun:
Hi =)

I'm making [extensions](https://github.com/OlegKrikun/commonmark-kotlinx-html) that rendering document into [kotlinx.html](https://github.com/Kotlin/kotlinx.html) stream, but `Renderer` interface accept only `Appendable`/`String` as result type.

This PR add ability to render into something other than string.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/Renderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/Renderer.java
@@ -2,7 +2,7 @@ package org.commonmark.renderer;
 
 import org.commonmark.node.Node;
 
-public interface Renderer {
+public interface Renderer<Output, Render> {
 
     /**
      * Render the tree of nodes to output.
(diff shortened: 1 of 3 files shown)


robinst:
Hey, can you point to the code where you actually use/implement the changed interface? I'm not sure how adding this abstraction is useful.

OlegKrikun:
Hey, currently I can't use this interface because my "extension" uses commonmark-java from maven repository. I've been planning to use it in [KotlinxHtmlRenderer](https://github.com/OlegKrikun/commonmark-kotlinx-html/blob/master/commonmark-kotlinx-html/src/main/kotlin/ru/krikun/commonmark/kotlinx/html/renderer.kt#L9). 

This change isn't really necessary, but it will allow to use Renderer interface instead of  KotlinxHtmlRenderer in client code.

robinst:
Closing this because it would break compatibility if someone was already using the existing `Renderer`, and not sure it's worth it.

PR #160, opened 2019-11-08
Fix possible stackoverflow error for regular expressions inside parser.

lehvolk:
Regular expressions used by inline parser may fail with Stackoverflow error. I made `match` method fail safe in this case. 

I will appreciate shipping this fix.


commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
```diff
@@ -243,17 +243,21 @@ private Node parseInline(Node previous) {
...
+            Matcher matcher = re.matcher(input);
+            matcher.region(index, input.length());
+            boolean m = matcher.find();
+            if (m) {
+                index = matcher.end();
+                return matcher.group();
+            } else {
+                return null;
+            }
+        } catch (Error e) {
```
robinst: Task: Seems like a good idea in general, but can we catch the specific error type here instead?

commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
```diff
@@ -184,7 +184,7 @@ private Text text(String text) {
      * On success, return the new inline node.
      * On failure, return null.
      */
-    private Node parseInline(Node previous) {
+    protected Node parseInline(Node previous) {
```
robinst: Why change visibility of all these methods? Please undo for this PR.

commonmark/src/test/java/org/commonmark/test/SpecialInputTest.java
```diff
@@ -140,6 +140,11 @@ public void emphasisMultipleOf3Rule() {
...
...Name=\"List Table 4 Accent 1\"/> <w:xxx Locked=\"false\" Priority=\"50\" Name=\"List Table 5 Dark Accent 1\"/>  <w:xxx Locked=\"false\" Priority=\"52\" Name=\"List Table 7 Colorful Accent 1\"/> <w:xxx Locked=\"false\" Priority=\"46\" Name=\"List Table 1 Light Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"47\" Name=\"List Table 2 Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"48\" Name=\"List Table 3 Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"49\" Name=\"List Table 4 Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"50\" Name=\"List Table 5 Dark Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"51\" Name=\"List Table 6 Colorful Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"52\" Name=\"List Table 7 Colorful Accent 2\"/> <w:xxx Locked=\"false\" Priority=\"46\" Name=\"List Table 1 Light Accent 3\"/> <w:xxx Locked=\"false\" Priority=\"47\" Name=\"List Table 2 Accent 3\"/> <w:xxx Locked=\"false\" Priority=\"48\" Name=\"List Table 3 Accent 3\"/> <w:xxx Locked=\"false\" Priority=\"49\" Name=\"List Table 4 Accent 3\" /> <w:xxx Locked=\"false\" Priority=\"50\" Name=\"List Table 5 Dark Accent 3\"/><w:xxx Locked=\"false\" Priority=\"51\" Name=\"List Table 6 Colorful Accent 3\"/></xml>");
```
robinst: Which regex in particular leads to the error? If possible we should also try to fix it.

lehvolk:
Please find remarks fixed.

Regarding methods visibility thought that it will be useful to have a way for further workarounds for parser plugged in current implementation. The way when you need to copy paste whole code from `InlineParserImpl` is not acceptable.

commonmark/src/test/java/org/commonmark/test/SpecialInputTest.java
lehvolk: It's `org.commonmark.internal.InlineParserImpl#HTMLTAG` Java's builtin regex support has problems with regexes containing repetitive alternative paths like HTMLTAG. Thought that this is a root case. Not sure it can be improved somehow to prevent such situations.

robinst:
Hey, thanks for addressing the feedback!

I'll look at this after I get back from leave. In the meantime, can you check if you need to sign the CLA, see here?:

https://github.com/atlassian/commonmark-java/blob/master/CONTRIBUTING.md#cla

robinst:
Merged, thanks!

robinst:
Released in 0.13.1: https://github.com/atlassian/commonmark-java/blob/master/CHANGELOG.md#0131---2019-11-25

PR #161, opened 2020-01-13
Add option for stripping unsafe URLs

### LINKED ISSUE #155: HTML renderer: Add option for stripping unsafe URLs in links and images
Steps to reproduce the problem (provide example input):
Unlike the [commonmark.js](https://github.com/commonmark/commonmark.js) plugin where the option `safe: true` can be used, the commonmark-java does not offer this option and thus XSS attacks can be done.

example can be as simple as this javascript alert when you add it to the markdown:

```
[Click this!](javascript:alert('message');)
```

Expected behavior:
Javascript should be removed from or escaped in the markdown.

Actual behavior:
Javascript will be executed.

VandorpeDavid:
Fixes #155

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
@@ -185,7 +185,7 @@ public class CoreHtmlNodeRenderer extends AbstractVisitor implements NodeRendere
         String altText = altTextVisitor.getAltText();
 
         Map<String, String> attrs = new LinkedHashMap<>();
-        if(context.shouldBeSafe()) {
+        if (context.shouldBeSafe()) {
             url = context.urlSanitizer().sanitizeImageUrl(url);
         }
         attrs.put("src", url);
--- a/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
@@ -143,7 +143,7 @@ public class CoreHtmlNodeRenderer extends AbstractVisitor implements NodeRendere
         Map<String, String> attrs = new LinkedHashMap<>();
         String url = link.getDestination();
 
-        if (context.shouldBeSafe()) {
+        if (context.shouldSanitizeUrls()) {
             url = context.urlSanitizer().sanitizeLinkUrl(url);
             attrs.put("rel", "nofollow");
         }
@@ -185,7 +185,7 @@ public class CoreHtmlNodeRenderer extends AbstractVisitor implements NodeRendere
         String altText = altTextVisitor.getAltText();
 
         Map<String, String> attrs = new LinkedHashMap<>();
-        if (context.shouldBeSafe()) {
+        if (context.shouldSanitizeUrls()) {
             url = context.urlSanitizer().sanitizeImageUrl(url);
         }
         attrs.put("src", url);
--- a/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
@@ -178,7 +178,7 @@ public class CoreHtmlNodeRenderer extends AbstractVisitor implements NodeRendere
 
     @Override
     public void visit(Image image) {
-        String url = context.encodeUrl(image.getDestination());
+        String url = image.getDestination();
 
         AltTextVisitor altTextVisitor = new AltTextVisitor();
         image.accept(altTextVisitor);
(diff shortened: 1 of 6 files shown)


commonmark/src/main/java/org/commonmark/renderer/html/HtmlNodeRendererContext.java
```diff
@@ -44,4 +46,16 @@
      * @return whether HTML blocks and tags should be escaped or not
      */
     boolean shouldEscapeHtml();
+
+    /**
+     *
+     * @return true if the {@link UrlSanitizer} should be used.
+     */
+    boolean shouldBeSafe();
```
robinst: Let's use a more descriptive name for this, how about `shouldSanitizeUrls`?

commonmark/src/main/java/org/commonmark/renderer/html/DefaultUrlSanitizer.java
```diff
@@ -0,0 +1,83 @@
...
+                    | (1L << '\u000c')
+                    | (1L << '\r');
+
+
+    public DefaultUrlSanitizer() {
+        this(new ArrayList<String>() {{
+            add("http");
+            add("https");
+            add("mailto");
+        }});
```
robinst: Use `Arrays.asList` instead.

commonmark/src/main/java/org/commonmark/renderer/html/DefaultUrlSanitizer.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Collection;
+
+/**
+ *
+ * Allows http, https and mailto protocols for url.
+ * Also allows protocol relative urls, and relative urls.
+ * Implementation based on https://github.com/OWASP/java-html-sanitizer/blob/f07e44b034a45d94d6fd010279073c38b6933072/src/main/java/org/owasp/html/FilterUrlByProtocolAttributePolicy.java
+ */
+public class DefaultUrlSanitizer implements UrlSanitizer {
+    private Collection<String> protocols;
```
robinst: Use a `Set` instead?

commonmark/src/main/java/org/commonmark/renderer/html/DefaultUrlSanitizer.java
```diff
@@ -0,0 +1,83 @@
...
+    public DefaultUrlSanitizer() {
+        this(new ArrayList<String>() {{
+            add("http");
+            add("https");
+            add("mailto");
+        }});
+    }
+
+    public DefaultUrlSanitizer(Collection<String> protocols) {
+        this.protocols = protocols;
```
robinst: Let's make a defensive copy of the passed in Collection here using `new HashSet<String>(protocols)`

commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
```diff
@@ -178,6 +185,9 @@ public void visit(Image image) {
         String altText = altTextVisitor.getAltText();
 
         Map<String, String> attrs = new LinkedHashMap<>();
+        if(context.shouldBeSafe()) {
```
robinst: Nit: missing space after if

commonmark/src/main/java/org/commonmark/renderer/html/DefaultUrlSanitizer.java
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+        if (i == 0 && n == s.length()) {
+            return s;
+        }
+        return s.substring(i, n);
+    }
+
+    private boolean isHtmlSpace(int ch) {
+        return ch <= 0x20 && (HTML_SPACE_CHAR_BITMASK & (1L << ch)) != 0;
```
robinst: Hmm, is this better than a switch?:

```
switch (ch) {
    case ' ':
    ...
}
```

commonmark/src/main/java/org/commonmark/renderer/html/HtmlNodeRendererContext.java
VandorpeDavid: Are you suggesting the property should be renamed (eg: `builder.sanitizeUrls(true)` instead of `builder.save(true)`), or just this function?

commonmark/src/main/java/org/commonmark/renderer/html/DefaultUrlSanitizer.java
VandorpeDavid: Probably not, this was copied from the OWASP implementation.

commonmark/src/main/java/org/commonmark/renderer/html/HtmlNodeRendererContext.java
robinst: All the related methods, yeah.

commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
```diff
@@ -178,6 +185,9 @@ public void visit(Image image) {
         String altText = altTextVisitor.getAltText();
 
         Map<String, String> attrs = new LinkedHashMap<>();
+        if (context.shouldSanitizeUrls()) {
+            url = context.urlSanitizer().sanitizeImageUrl(url);
```
robinst: I noticed that for `Links`, you pulled out the `encodeUrl` to happen *after* sanitizing. But here it happens before (expand the context above).

They should be consistent, and I think it makes sense to have `encodeUrl` be after sanitizing.

robinst:
Thanks for the contribution! ❤️

I'll also add a note to the readme about this option, and then do a release sometime this week.

Issue #113, opened 2017-11-10
Make inline parsing extensible


robinst:
Similar to how custom block parser can be registered today, it would be good to make inline parsing extensible as well. One use would be to improve autolinking, see #112.

The way this could work:

* When we match a character during inline parsing, the last fallback is to treat it as text. Before we do that, if there are custom inline parsers, we could call each one (with some priority maybe?) and give them the opportunity to handle that.
* In a custom inline parser, it could parse the input and return/add Nodes. It could also advance the current parsing position. So inline parsing needs to expose some API (or the custom inline parser would return instructions).

To support autolinking, a custom inline parser would also need to be able to get the text before the matching character, probably all the plain text characters up to that point. It would also need a way to detect if any brackets are active or not ("are we inside a link?").

robinst:
I'm pretty sure I'm not gonna pick this up in the near future. But if someone else wants to have a go at it, I'm willing to mentor. Help wanted :).

HiuKwok:
Hi @robinst, I am willing to work on it. However as a newbie of both codebase and Commonmark spec. Just wanna clarify several points first to make sure I m on the same page as you guys.   

* By make inline parsing extendable, do you mean first refactor `inlineParserImpl` one central class into multiples `inlineParserImpl` for all inline components that listed on [spec](https://spec.commonmark.org/0.28/). So each of the components impl can have chance to parser the content in a whole line lv instead of how it handle at the moment where `InlinePraserImpl.parseInline()` ?
* Also by this mean, interface `BlockParser.parserInlines()` would have a new signature which take a list of InlineParser as a input?

HiuKwok:
Hi everyone, several commit have been made for let Both parser and doc parser class to accept multiple instances of InlineParser. 
01f41492867e30ac067928989d1aaed679e9a735
9b28f867d156996ce5281301f7cd9bc18296d7ea
e145fc521cd915a2a2cc10e09f95d65141636bcb 

My next thought would be alter InlineParser interface to make it return a parsing progress object which is similar to `BlockContinue` class, so when block parser attempt to parser it content via lists of InlineParser. Every inlineParser would be able to pick where last InlineParser left up to. 

Am I on the right track? Plz advise

robinst:
> Every inlineParser would be able to pick where last InlineParser left up to.

Yes, that's the rough idea (similar to block parsers).

[...]

PR #164, opened 2020-03-23
An extension for adding attributes to nodes, specifically for adding height and width to image nodes

dohertyfjatl:
* style elements are given as key=value pairs inside curly braces, for example:
    `text{height=5 width=6}`
  after the node to which they apply
* only allow height and width as the supported set of style keys
* only support styles for image nodes


dohertyfjatl:
I was trying to find a way of supporting height and width for images.

They are not officially part of the common mark spec, so I wanted to add them as an extension. 

I found that one way it could be done relatively easily is by using curly braces as the delimiters and then retrieving the styles as key=value pairs.

Although this approach could be used fairly generically for adding style details to nodes, I added a couple of restrictions:
* only allow "width" and "height" as the keys
* only allow styles for image nodes

commonmark-ext-styles/src/test/java/org/commonmark/ext/styles/StylesTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.commonmark.Extension;
+import org.commonmark.parser.Parser;
+import org.commonmark.renderer.html.HtmlRenderer;
+import org.commonmark.testutil.RenderingTestCase;
+import org.junit.Test;
+
+import java.util.Collections;
+import java.util.Set;
+
+public class StylesTest extends RenderingTestCase {
```
bplump: Do you need any negative tests here? Like a check that `"This is some text with random styles afterwards {width=100px}"` just renders as text?

bplump:
Given the lack of standardization on image styling, this approach seems as good as any.

When I looked around at implementations, Kramdown's approach seemed to get mentioned more than most, which is almost the same but starts with `{:`. But... ¯\_(ツ)_/¯ 

I'd love Commonmark to define a standard but I'm guessing they don't feel the need because "if you need that much customization, just use HTML tags"... which is fair.

commonmark-ext-styles/src/test/java/org/commonmark/ext/styles/StylesTest.java
dohertyfjatl: Makes sense. Done.

robinst:
There's a long discussion about image (and generic) attributes here: https://talk.commonmark.org/t/consistent-attribute-syntax/272

Unfortunately there's no outcome yet. It does look like people are leaning towards `text{height=5 width=6}`. Note that's also what pandoc (written by jgm, the author of the CommonMark spec) has: https://pandoc.org/MANUAL.html#extension-link_attributes

You can play around with it here, some examples:
* https://pandoc.org/try/?text=!%5B%5D(file.jpg)%7Bwidth%3D50%7D&from=markdown&to=&standalone=0
* https://pandoc.org/try/?text=!%5B%5D(file.jpg)%7Bwidth%3D50+*invalid*%7D&from=markdown&to=&standalone=0

I'm fine with having this as an extension as you implemented it. I'll comment on the code.

README.md
```diff
@@ -319,6 +319,17 @@ document start here
...
 Use class `YamlFrontMatterExtension` in artifact `commonmark-ext-yaml-front-matter`. To fetch metadata, use `YamlFrontMatterVisitor`.
 
+### Styles
+
+Adds support for specifying style details (specifically height and width) for images.
+
+The style elements are given as key=value pairs inside curly braces after the node to which they apply, for example:
+```
+![text](/url.png){height=5 width=6}
+```
```
robinst: Can you add the HTML this produces?

README.md
```diff
@@ -319,6 +319,17 @@ document start here
...
+
+Use class `StylesExtension` in artifact `commonmark-ext-styles`.
```
robinst: The name is a bit too generic I think. How about `commonmark-ext-image-attributes` instead? Attributes is also what it's called in the discussions and in pandoc: https://pandoc.org/MANUAL.html#extension-link_attributes

If it ever gets included in the CommonMark spec, this will live in core and we can deprecate the extension.

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/Styles.java
```diff
@@ -0,0 +1,35 @@
+package org.commonmark.ext.styles;
+
+import org.commonmark.node.CustomNode;
+import org.commonmark.node.Delimited;
+
+/**
+ * A styles node containing text and other inline nodes as children.
+ */
+public class Styles extends CustomNode implements Delimited {
```
robinst: Rename to `ImageAttributes`?

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesAttributeProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void setAttributes(Node node, String tagName, final Map<String, String> attributes) {
+        if (node instanceof Image) {
+            node.accept(new AbstractVisitor() {
+                @Override
+                public void visit(CustomNode node) {
+                    if (node instanceof Styles) {
+                        Styles styles = (Styles) node;
+                        for (String s: styles.getStyles().split(" ")) {
```
robinst: Style nit: missing space before `:`

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesDelimiterProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+        return 1;
+    }
+
+    @Override
+    public void process(Text opener, Text closer, int delimiterCount) {
+        StringBuilder styleString = new StringBuilder();
+        Node tmp = opener.getNext();
+        while (tmp != null && tmp != closer) {
+            Node next = tmp.getNext();
+            if (tmp instanceof Text) {
```
robinst: Pandoc's behavior is to fall back to text if it can't be parsed, e.g.:
* https://pandoc.org/try/?text=!%5B%5D(file.jpg)%7Bwidth%3D50+*test*%7D&from=markdown&to=&standalone=0
* https://pandoc.org/try/?text=!%5B%5D(file.jpg)%7Bwidth%3D50+baa%7D&from=markdown&to=&standalone=0

Can you implement that here (and add a test?). We'd need to do some more parsing here and bail for other node types.

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesDelimiterProcessor.java
```diff
@@ -0,0 +1,51 @@
+package org.commonmark.ext.styles.internal;
+
+import org.commonmark.ext.styles.Styles;
+import org.commonmark.node.Node;
+import org.commonmark.node.Text;
+import org.commonmark.parser.delimiter.DelimiterProcessor;
+import org.commonmark.parser.delimiter.DelimiterRun;
+
+public class StylesDelimiterProcessor implements DelimiterProcessor {
```
robinst: Implementing it as a delimiter processor + attribute provider is clever :)! I think if we supported extending inline parsing (see #113), this would probably a good use for it. But we don't, so a delimiter processor seems nice.

The only problem I see is that it means `{` can't be used by other delimiter processors anymore. I would probably mention that in the README.

Oh, one more thing, does having spaces inside the braces work?: `{ width=50 }`

README.md
dohertyfjatl: Added.

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesDelimiterProcessor.java
dohertyfjatl: Added note to the README.

dohertyfjatl: > Oh, one more thing, does having spaces inside the braces work?: `{ width=50 }`

Unfortunately no, as it seems that `InlineParserImpl` will not recognise the opening delimiter if it is followed by whitespace:
https://github.com/atlassian/commonmark-java/blob/cb4bee61046e21a0bef4ab96860334d1f089fba7/commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java#L703,L715

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesDelimiterProcessor.java
```diff
@@ -30,22 +37,95 @@ public int getDelimiterUse(DelimiterRun opener, DelimiterRun closer) {
...
+            List<Node> toUnlink = new ArrayList<>();
+
+            StringBuilder styleString = new StringBuilder();
+            Node tmp = opener.getNext();
+            while (tmp != null && tmp != closer) {
+                Node next = tmp.getNext();
+                // Only Text nodes can be used for attributes
+                if (tmp instanceof Text) {
+                    String styles = ((Text) tmp).getLiteral();
+                    for (String s : styles.split("\\s+")) {
```
dohertyfjatl: Split on `\s+` to support the case where more than one space is used to separate the attributes

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesDelimiterProcessor.java
```diff
@@ -30,22 +37,95 @@ public int getDelimiterUse(DelimiterRun opener, DelimiterRun closer) {
...
+                    // The styles node is added as a child of the node to which the styles apply.
+                    Node nodeToStyle = opener.getPrevious();
+                    nodeToStyle.appendChild(styles);
+                }
+                return;
+            }
+        }
+
+        // If we got here then the attributes cannot be applied, so fallback to leaving the text unchanged.
+        // Need to add back the opening and closing characters (which are removed elsewhere).
```
dohertyfjatl: The opening and closing characters get removed elsewhere, so we need to add them back here.

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesDelimiterProcessor.java
dohertyfjatl: Added support for falling back to text, and added/updated tests to confirm this behaviour.

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/internal/StylesAttributeProvider.java
dohertyfjatl: Fixed.

commonmark-ext-styles/src/main/java/org/commonmark/ext/styles/Styles.java
dohertyfjatl: Done.

README.md
dohertyfjatl: Renamed to `commonmark-ext-image-attributes`.

commonmark-ext-image-attributes/src/main/java/org/commonmark/ext/image/attributes/internal/ImageAttributesDelimiterProcessor.java
```diff
@@ -0,0 +1,131 @@
...
+            StringBuilder attributesString = new StringBuilder();
+            Node tmp = opener.getNext();
+            while (tmp != null && tmp != closer) {
+                Node next = tmp.getNext();
+                // Only Text nodes can be used for attributes
+                if (tmp instanceof Text) {
+                    String attributes = ((Text) tmp).getLiteral();
+                    for (String s : attributes.split("\\s+")) {
+                        String[] attribute = s.split("=");
+                        // Check if the attribute is in SUPPORTED_STYLES.
```
bplump, line 55:
```suggestion
                        // Check if the attribute is in SUPPORTED_ATTRIBUTES.
```

bplump: Though... not really sure if the comment is necessary.

commonmark-ext-image-attributes/src/test/java/org/commonmark/ext/image/attributes/ImageAttributesTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @Test
+    public void unsupportedStyleNamesAreLeftUnchanged() {
+        assertRendering("![text](/url.png){j=502 K=101 img=2 url=5}",
+                "<p><img src=\"/url.png\" alt=\"text\" />{j=502 K=101 img=2 url=5}</p>\n");
+        assertRendering("![foo](/url.png){height=3 invalid}\n",
+                "<p><img src=\"/url.png\" alt=\"foo\" />{height=3 invalid}</p>\n");
+        assertRendering("![foo](/url.png){height=3 *test*}\n",
+                "<p><img src=\"/url.png\" alt=\"foo\" />{height=3 <em>test</em>}</p>\n");
+    }
+
```
bplump, line 62: This test fails with an `ArrayIndexOutOfBoundsException` when rendering. I think the delimiter processor should ensure each key has a value.

```suggestion

    @Test
    public void styleWithNoValueIsIgnored() {
        assertRendering("![text](/url.png){height}",
                "<p><img src=\"/url.png\" alt=\"text\" />{height}</p>\n");
    }
        
```

commonmark-ext-image-attributes/src/main/java/org/commonmark/ext/image/attributes/internal/ImageAttributesDelimiterProcessor.java
dohertyfjatl: I think you're right that it's not necessary, so I've removed it.

commonmark-ext-image-attributes/src/test/java/org/commonmark/ext/image/attributes/ImageAttributesTest.java
dohertyfjatl: I've added a check on `attribute.length` to `ImageAttributesAttributeProvider` and
  `ImageAttributesDelimiterProcessor`, and also included your new test.

Thanks for pointing that out!

commonmark-ext-image-attributes/src/main/java/org/commonmark/ext/image/attributes/internal/ImageAttributesDelimiterProcessor.java
```diff
@@ -0,0 +1,131 @@
...
+        closer.getParent().appendChild(new Text("" + getClosingCharacter()));
+    }
+
+    /**
+     * Check that the attributes can be applied to the previous node.
+     * @param opener the text node that contained the opening delimiter
+     * @param closer the text node that contained the closing delimiter
+     * @return true if the previous node is an Image and the attributes are in the set of {@link #SUPPORTED_ATTRIBUTES}
+     */
+    private boolean canApply(Text opener, Text closer) {
```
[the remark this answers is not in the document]
dohertyfjatl: Removed.

commonmark-ext-image-attributes/src/main/java/org/commonmark/ext/image/attributes/ImageAttributes.java
```diff
@@ -0,0 +1,35 @@
...
+import org.commonmark.node.Delimited;
+
+/**
+ * A node containing text and other inline nodes as children.
+ */
+public class ImageAttributes extends CustomNode implements Delimited {
+
+    private final String attributes;
+
+    public ImageAttributes(String attributes) {
```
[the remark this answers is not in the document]
dohertyfjatl: I think you're right, so I've made this change now to use a `Map<String, String>`.

commonmark-ext-image-attributes/src/main/java/org/commonmark/ext/image/attributes/internal/ImageAttributesAttributeProvider.java
```diff
@@ -0,0 +1,40 @@
...
+        if (node instanceof Image) {
+            node.accept(new AbstractVisitor() {
+                @Override
+                public void visit(CustomNode node) {
+                    if (node instanceof ImageAttributes) {
+                        ImageAttributes imageAttributes = (ImageAttributes) node;
+                        Map<String, String> attributesMap = imageAttributes.getAttributes();
+                        for (String key : attributesMap.keySet()) {
+                            attributes.put(key, attributesMap.get(key));
+                        }
```
robinst, line 32: Can you change this to:

```suggestion
                        for (Map.Entry<String, String> entry : imageAttributes.getAttributes().entrySet()) {
                            attributes.put(entry.getKey(), entry.getValue());
                        }
```

commonmark-ext-image-attributes/src/main/java/org/commonmark/ext/image/attributes/internal/ImageAttributesDelimiterProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        // If we got here then the attributes cannot be applied, so fallback to leaving the text unchanged.
+        // Need to add back the opening and closing characters (which are removed elsewhere).
+        if (opener.getPrevious() == null) {
+            opener.getParent().prependChild(new Text("" + getOpeningCharacter()));
+        } else {
+            opener.getPrevious().insertAfter(new Text("" + getOpeningCharacter()));
+        }
+        closer.getParent().appendChild(new Text("" + getClosingCharacter()));
```
robinst, line 98: (Note to self: If I ever rework the delimiter interface, it might make sense to leave removing of the delimiters to the implementation instead of doing it for all delimiters. Then this code here would not be necessary..)

robinst:
Merged now, thanks!

jk1:
@dohertyfjatl thanks a ton for implementing this.
Is it by chance published to a maven repo already?

robinst:
I want to get the other open PR merged first. Then I'll publish a release.

PR #168, opened 2020-05-12
Improve url sanitization by unescaping html entities.

ysun-atl:
This pull request:
- Unescapes HTML from the input url before filtering the protocols to prevent XSS via HTML hex/decimal references using `commons-text`
- Adds tests for potential XSS scenarios
- Rearranged some code

### DIFF
--- a/commonmark/pom.xml
+++ b/commonmark/pom.xml
@@ -13,6 +13,11 @@
 
     <dependencies>
         <dependency>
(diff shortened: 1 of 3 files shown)


robinst:
Note that this was not necessary because rendering `&#x6A&#x61` with HtmlRenderer would result in `&amp;#x6A&amp;#x61` (`&` escaped), so no XSS there.

PR #180, opened 2020-07-27
Inline parsing overhaul WIP

robinst:
The end goal is to extract all current inline parsers, and then allow customizing inline parsing via the same infrastructure (similar to the current block parsing).

A stretch goal is to move away from the parser input having to be a single contiguous String. This could speed up parsing and allow keeping track of source positions (line and column indexes) for inline nodes.

Still todo:
* [x] Replace remaining direct usages of `index` and `match` in `InlineParserImpl` with `Scanner`
* [x] See if we can change `input` from `String` to `List<CharSequence>`
* [ ] Clean up internal API

(Allowing extension of inline parsers can be done in a separate PR, but it will be much simpler now.)


commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
```diff
@@ -163,44 +143,48 @@ public void parse(String content, Node block) {
...
-        char c = peek();
+        Scanner scanner = scanner();
+        char c = scanner.peek();
         if (c == '\0') {
             return null;
         }
 
+        List<InlineContentParser> inlineParsers = this.inlineParsers.get(c);
+        if (inlineParsers != null) {
+            for (InlineContentParser inlineParser : inlineParsers) {
```
samukce, line 175: hmm I applied a similar idea here https://github.com/atlassian/commonmark-java/pull/178/files#diff-abb5406ad2ea9b67e29591c9ffb7d35bR45

One challenge that I saw was in case we do have more than one `InlineContentParser` using the same start symbol. We don't have how to know in the current position which one will find the "end character" to decide which one is the valid one. That's why I kept one "InlineContentParser" active list and one array where I could notice which node could be fully parsed.

I can see that we have one example which has the same start symbol
`this.inlineParsers.put('<', Arrays.asList(new AutolinkInlineParser(), new HtmlInlineParser()));` 
For `AutolinkInlineParser` we need from the start symbol until the close `>` but we may not have, right? Thinking when we add here the extension of how to handle this in case the client code also adds a third InliineParser for `<`.. Any idea?

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
```diff
@@ -0,0 +1,115 @@
+package org.commonmark.internal.inline;
+
+import org.commonmark.internal.util.CharMatcher;
+import org.commonmark.internal.util.Parsing;
+
+public class Scanner {
```
samukce, line 6: This scanner class helped a lot the code readability. :)

commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
robinst: Not sure what you mean, can you explain with an example? With this, the first inline parser that returns a result will win. To decide, the inline parser can look at as much of the input as it wants (until the end of the block).

For custom inline parsers, they should probably come first, so they can modify the core syntax. In this case, if a custom inline parser for `<` was registered, it could do whatever it wants, before the core inline parsers would have a chance.

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
robinst: Yeah, I think so too :).

commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
samukce: Summaryzing what I tried to say is that the number of interactions will be multiplied by the number of (Node)Parsers.

Let me try to put the example below:

```
Taking the defaults parsers:
'<'
  AutolinkInlineParser -> scanner.find('>') > 0
    URI.matcher(text).matches()
    or
    EMAIL.matcher(text).matches()
   
  HtmlInlineParser
    Regex
    or
    to look for some "end character" in the line
    


Parsing the line below:
  Some dummy text < anybloh text here>
                  ^
                  -------------------^
                    AutolinkInlineParser,
                      return None
                  ^
                  -------------------^
                    HtmlInlineParser
                      return None    
``` 

if we add extensions for `<` we would have `(N * extensions start by '<' )` interacting in this line. But thinking better we may not have how to avoid it.


The model that I tried in that other PR it kind of a little bit different for that one that you took. Taking the same example the algorithm is:

```
Parsing the line below:
  Some dummy text < anybloh text here>
                  ^
                  AutolinkInlineParser/HtmlInlineParser active
                  -------------------^
                                     AutolinkInlineParser.isComplete()?
                                      return None;
                                     HtmlInlineParser.isComplete()?
                                      return None
```

Frankly speaking, I don't know which one is the best. It seems the approach that you took we can avoid holding multiple parsers watching the line at the same time. So, I'm with you. :)

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public char peek() {
+        if (index < lineLength) {
+            return line.charAt(index);
+        } else {
+            if (lineIndex < lines.size() - 1) {
+                return '\n';
+            } else {
+                // Don't return newline for end of last line
+                return '\0';
```
samukce, line 41: I can see we using this symbol in multiple places. What do you think to extract and also use the constant from Character class? `Character.MIN_VALUE`?

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
```diff
@@ -0,0 +1,187 @@
...
+                lineLength = line.length();
+            } else {
+                line = "";
+                lineLength = 0;
+            }
+            index = 0;
+        }
+    }
+
+    public boolean next(char c) {
```
samukce, line 83: While reading the code this `next` method makes me think a different action for this method. When I renamed to `isNext(...)` it more senses to me. what do you think?

commonmark/src/main/java/org/commonmark/internal/util/AsciiMatcher.java
```diff
@@ -0,0 +1,52 @@
...
+    public boolean matches(char c) {
+        return set.get(c);
+    }
+
+    public Builder newBuilder() {
+        return new Builder((BitSet) set.clone());
+    }
+
+    public static Builder builder() {
+        return new Builder(new BitSet());
```
samukce, line 23: The `Builder`class could have this boot. 
```java
return new Builder();
...
 private Builder() {
            this.set = new BitSet();
        }
```

commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private static boolean tryCdata(Scanner scanner) {
+        // spec: A CDATA section consists of the string <![CDATA[, a string of characters not including the string ]]>,
+        // and the string ]]>.
+
+        // Skip `[`
+        scanner.next();
+
+        if (scanner.next('C') && scanner.next('D') && scanner.next('A') && scanner.next('T') && scanner.next('A')
+                && scanner.next('[')) {
```
samukce, line 186: To have an overload `next(...)` method for multiple char could increase even more the readability here.

Something like
```java
    if (scanner.next('C', 'D', 'A', 'T', 'A', '[') {
   ....

    //Scanner.java
    public boolean next(char... c) {
        for (char character: c) {
            if (!isNext(character)){
                return false;
            }
        }
        return true;
    }

   //ScannerTest.java
    @Test
    public void testIsNextMultipleCharacter() {
        Scanner scanner = new Scanner(Collections.<CharSequence>singletonList("foo bar"), 0, 0);
        assertTrue(scanner.isNext('f', 'o', 'o', ' ', 'b', 'a', 'r'));
        assertFalse(scanner.isNext('r'));
    }
```

commonmark/src/main/java/org/commonmark/internal/inline/BackslashInlineParser.java
```diff
@@ -0,0 +1,35 @@
...
+        char next = scanner.peek();
+        if (next == '\n') {
+            scanner.next();
+            return ParsedInline.of(new HardLineBreak(), scanner.position());
+        } else if (ESCAPABLE.matcher(String.valueOf(next)).matches()) {
+            scanner.next();
+            return ParsedInline.of(new Text(String.valueOf(next)), scanner.position());
+        } else {
+            return ParsedInline.of(new Text("\\"), scanner.position());
+        }
```
samukce, line 33: Just to simplify those if/else and have less indentation. 

```suggestion
        if (next == '\n') {
            scanner.next();
            return ParsedInline.of(new HardLineBreak(), scanner.position());
        }

        if (ESCAPABLE.matcher(String.valueOf(next)).matches()) {
            scanner.next();
            return ParsedInline.of(new Text(String.valueOf(next)), scanner.position());
        } 

        return ParsedInline.of(new Text("\\"), scanner.position());
```

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
```diff
@@ -0,0 +1,187 @@
...
+            return lines.get(begin.lineIndex).subSequence(begin.index, end.index);
+        } else {
+            StringBuilder sb = new StringBuilder();
+
+            CharSequence firstLine = lines.get(begin.lineIndex);
+            sb.append(firstLine.subSequence(begin.index, firstLine.length()));
+            sb.append('\n');
+
+            // Lines between begin and end (we are appending the full line)
+            for (int line = begin.lineIndex + 1; line < end.lineIndex; line++) {
```
samukce, line 177: We do have an attribute in this line called `line` too. We may call here by simple `i,j,....`

commonmark/src/main/java/org/commonmark/internal/util/AsciiMatcher.java
```diff
@@ -0,0 +1,52 @@
...
+    public static class Builder {
+        private final BitSet set;
+
+        private Builder(BitSet set) {
+            this.set = set;
+        }
+
+        public Builder c(char c) {
+            if (c > 127) {
+                throw new IllegalArgumentException("Can only match ASCII characters");
```
samukce, line 35:
```suggestion
                throw new IllegalArgumentException("Can only match ASCII Printable Characters");
```

commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
robinst: Ah yeah, I know what you mean now. I think given that it works the same way as before this change, it should be ok. I like the model with this change, it's simple to understand and works the same way as block parsers.

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
robinst: Good idea, yeah! `Character.MIN_VALUE` would work, but I think something even more specific would be nicer like `Parsing.EOF` or `Parsing.END`.

commonmark/src/main/java/org/commonmark/internal/inline/Scanner.java
robinst: Yeah, it's really a `nextIf`. What I like about `next` is that it reads well in code like `if (scanner.next('x'))`.

With `isNext` it's not clear that it advances the scanner state to me.

commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java
robinst: Yeah, I was considering this before, maybe even a `next(String)` instead, which only advances if it finds the whole string.

commonmark/src/main/java/org/commonmark/parser/InlineParser.java
```diff
@@ -2,14 +2,17 @@
...
 public interface InlineParser {
 
     /**
-     * @param input the content to parse as inline
+     * @param lines the content to parse as inline
      * @param node the node to append resulting nodes to (as children)
      */
-    void parse(String input, Node node);
+    // TODO: Should we use a better type here, one that will be able to include source positions?
+    void parse(List<CharSequence> lines, Node node);
```
robinst, line 17: I'm happy with this PR now (for internal API, would still need some more cleanup to be public API), but I still want to spike what this type would look like with source positions, so that we can avoid breaking API twice.

PR #182, opened 2020-08-04
Removed test code duplication

leofernandesmo:
Undesirable duplication in tst code. 
Solutions are similar to those for normal code duplication as described by Fowler.
So, I applied extract method refactoring to remove duplication.


robinst:
Sure, thansk.

PR #184, opened 2020-08-31
Add GitHub actions

robinst:


robinst:
Travis is still used for testing on Android, but we can migrate that in a separate step.

PR #185, opened 2020-09-08
Support for including source spans on block nodes

### LINKED ISSUE #1: Add source position/maps to AST
Having source positions is a useful feature for editors, as it allows linking blocks between the source and the rendered output. commonmark.js supports it, see highlighted blocks in preview of [dingus](http://spec.commonmark.org/dingus/).

There is some code for adding source positions to blocks, but it's untested and not currently exposed.

robinst:
- Answer for "Where in the source input (line/column position and length) does this block come from?"
- Useful for things like editors that want to keep the input and rendered output scrolled to the same lines,
  or start editing on the block that was selected.
- Use `includeSourceSpans` on `Parser.Builder` to enable, read data with `Node.getSourceSpans`

Fixes #1 for block nodes. Will look into inline nodes too after #180, which might require more API changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+## Unreleased
+### Added
+- Support for including source spans on block nodes:
+  - Answer for "Where in the source input (line/column position and length) does this block come from?"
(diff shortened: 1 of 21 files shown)


Mi-La:
@robinst Hi! Just a note. Knowing the position in source is also very helpful when you need to report warnings/errors.

E.g. I'm using HTMLRenderer and I'm trying to verify that links lead to "somewhere", and when the linked resource is missing, I would like to report a warning. But currently I've no chance to get source location from e.g. Link node. Looking forward for the new release! :-). Do you have any plans already?

Or is there any other workaround how to get position in source markdown? (I'm using lastest release 0.15.2).

PR #191, opened 2020-11-18
add dot match in metadata

javaor:
Sometimes we will have key like **a.b** in metadata, wish it will be support.

e.g. Below format:
```
---
ms.author: author
ms.date: yyyy/MM/dd
---
```


robinst:
Hey! Change makes sense to me, as YAML also supports dots within plain (unquoted) keys.

Can you add a test case for this though? Should be simple to add, see `YamlFrontMatterTest`.

robinst (changes requested):
Please add a test

javaor:
Due to the change hasn't been merged, so the test case failed I think.

robinst:
Thanks! Will be in the next release 🔜

PR #197, opened 2020-12-21
Re-enable android tests

noties:
Hello,

At first I decided to remove `commonmark-android-test` folder all-together. As I find that the requirement to run on MacOS and launching emulator in a headless mode is a bit too heavy for a project that doesn't have Android as the primary target. I came up to this after discovering the [Github action to launch the emulator](https://github.com/marketplace/actions/android-emulator-runner/). 

But then I thought, couldn't Lint, the tool that is bundled with Android SDK distribution and which highlights different errors in IDE be used here? As it turns out it can. And it has no overhead of launching an emulator and can be run from a command line on any platform. I've set the most crucial for compatibility Lint issues as `fatal`, so a build task would fail if any of the issues occur. It immediately provided feedback and brought attention to issues already present in the code base. Somehow they did not reveal themself earlier. Most likely because majority of Android projects had already migrated to Java 8 and desugarring (provided by Android Gradle plugin), or have minimum SDK level increased past 16.

But Lint has one downside - it operates only on sources and cannot process project dependencies. `commonmark-java` has one dependency - [autolink](https://github.com/robinst/autolink-java) and currently it is not processed to ensure Android compatibility. Maybe a similar check can be additionally executed there, or project can be added as a submodule (maybe only for Android tests) or maybe there is a Gradle plugin that allows adding a dependency from VC (I did not find any that allows mixing different build systems - Gradle and Maven in our case).

Looking forward to your feedback 🙌 

BTW, build checks are failing because there are issues related to Android compatibility, which I think is the main goal of requested functionality


noties:
It seems that build checks are not triggered due to recent namespace migration, a job should've failed with output similar to [the one in my fork](https://github.com/noties/commonmark-java/actions/runs/435676967)

robinst:
Oh this is nice, I like that we don't have to have a complex build setup anymore. Thanks for working on this!

On the compatibility problems: Looks like we could raise the minimum API level to 19 to fix the issues, do you think that's reasonable or would it be good to stay on 16?

autolink: Yeah would be good to have a similar build running there. I'll look into it when I have time.

I think the reason the build is not running might be because `on.pull_request` is missing?: https://github.com/commonmark/commonmark-java/blob/f99163e6a2e660e0c4fc3b008391e44ef8b3422a/.github/workflows/ci.yml#L5

noties:
Let's update minimum SDK level, 19 is good - it still covers absolute majority of devices (98.1%).

I also updated the `ci.yml` file, so all checks are executed for pull requests

robinst:
Merged, thanks so much!

PR #244, opened 2021-09-05
Fix tables with leading/trailing header pipes and trailing spaces

bplump:
This is another table rendering issue introduced in 0.16.1. It's extremely obscure (found by running 0.18.0 against millions of markdown comments and comparing against 0.15.2) but I'm hoping the fix is simple enough that we can maintain compatibility with 0.15.2. 🤞 

It's probably easiest to look at the added test cases - note the space at the end of the header line in the first test. When the header row has leading/trailing pipes and the trailing pipe is followed by whitespace, the table fails to render. (Trailing whitespace on the separator line or content lines works fine; trailing whitespace when the header line doesn't have a trailing pipe also works fine)

[The 0.15.2 code would trim each line at the start of `split()`](https://github.com/commonmark/commonmark-java/blob/commonmark-parent-0.15.2/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java#L104). Without the trim, [`split()` adds the whitespace after the trailing pipe as another header cell](https://github.com/commonmark/commonmark-java/blob/commonmark-parent-0.18.0/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java#L155). As a result, [`tryStart` thinks there are more header cells than columns](https://github.com/commonmark/commonmark-java/blob/commonmark-parent-0.18.0/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java#L257) and stops trying to parse a table.

The current code [already removes any leading whitespace and any leading pipe](https://github.com/commonmark/commonmark-java/blob/commonmark-parent-0.18.0/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java#L125). If and only if there's a leading pipe, the fix does the same for the trailing pipe.

### DIFF
--- a/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java
+++ b/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/internal/TableBlockParser.java
@@ -122,14 +122,22 @@ public class TableBlockParser extends AbstractBlockParser {
     private static List<SourceLine> split(SourceLine line) {
         CharSequence row = line.getContent();
         int nonSpace = Parsing.skipSpaceTab(row, 0, row.length());
-        int cellStart = row.charAt(nonSpace) == '|' ? nonSpace + 1 : nonSpace;
+        int cellStart = nonSpace;
+        int cellEnd = row.length();
+        if (row.charAt(nonSpace) == '|') {
+            // This row has leading/trailing pipes - skip the leading pipe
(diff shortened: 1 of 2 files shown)


robinst:
> found by running 0.18.0 against millions of markdown comments and comparing against 0.15.2

Love it!

Thanks for the fix, I'll do a release with this shortly.

PR #251, opened 2022-01-21
Change license URL to "https"

### LINKED ISSUE #250: Change BSD-2-Clause license URL in the POM from "http" to "https" to conform to SPDX
We're using a [Gradle plugin](https://github.com/cashapp/licensee) that helps prevent inadvertent transitive dependencies with licenses we haven't explicitly allowlisted. Since Commonmark is licensed under BSD-2-Clause, it should be possible to allowlist it by the license's [SPDX](https://github.com/spdx/license-list-XML/blob/master/src/BSD-2-Clause.xml):

```
allow('BSD-2-Clause')
```

This doesn't work however as the license URL, [as indicated in the POM](https://github.com/commonmark/commonmark-java/blob/main/pom.xml#L267), is an "http" URL, while the SPDX spec requires it to be "https".

A simple fix is to change the URL to "https", I'll be happy to send a PR if there are no objections.

Egorand:
Fixes #250

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -264,7 +264,7 @@
     <licenses>
         <license>
             <name>BSD 2-Clause License</name>
-            <url>http://opensource.org/licenses/BSD-2-Clause</url>
+            <url>https://opensource.org/licenses/BSD-2-Clause</url>
             <distribution>repo</distribution>
         </license>
     </licenses>


robinst:
Thanks!

Issue #255, opened 2022-03-01
Dangling pipe breaks table block parser


jk1:
Steps to reproduce the problem (provide example Markdown if applicable):

```
Abc|Def
---|---
1|2
|
```

Expected behavior:

```
<table>
<thead>
<tr>
<th>Abc</th>
<th>Def</th>
</tr>
</thead>
<tbody>
<tr>
<td>1</td>
<td>2</td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</tbody>
</table>
```

Actual behavior:

```
java.lang.StringIndexOutOfBoundsException: String index out of range: 1

	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:48)
	at java.base/java.lang.String.charAt(String.java:709)
	at org.commonmark.ext.gfm.tables.internal.TableBlockParser.split(TableBlockParser.java:137)
	at org.commonmark.ext.gfm.tables.internal.TableBlockParser.parseInlines(TableBlockParser.java:80)
	at org.commonmark.internal.DocumentParser.processInlines(DocumentParser.java:475)
	at org.commonmark.internal.DocumentParser.finalizeAndProcess(DocumentParser.java:522)
	at org.commonmark.internal.DocumentParser.parse(DocumentParser.java:121)
	at org.commonmark.parser.Parser.parse(Parser.java:70)
	at org.commonmark.ext.gfm.tables.TablesTest.render(TablesTest.java:789)
	at org.commonmark.testutil.RenderingTestCase.assertRendering(RenderingTestCase.java:10)
	at org.commonmark.ext.gfm.tables.TablesTest.tableWithDanglingPipe(TablesTest.java:713)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
[...]

PR #256, opened 2022-03-01
#255 handle dangling pipes

jk1:


robinst:
Hey, thanks for looking into this. Looking at what GitHub does (which we're trying to mirror with the extension), it renders this:

```
Abc|Def
---|---
1|2
|
```

Like this:

Abc|Def
---|---
1|2
|

---

See how `|` is not actually part of the table anymore? Can you come up with a fix that handles it the same way?

robinst:
Hey @jk1 just checking in, have you seen my comment?

robinst:
Looked at it myself, the fix is a bit more involved, will raise a PR soon.

robinst:
PR: https://github.com/commonmark/commonmark-java/pull/268

jk1:
Hi @robinst thank you for the followup. I have seen the your comment back then, but didn't manage to implement what's requested. Things got wild really fast and was forced to flee my home country. Thanks a ton for getting back to this and providing a proper fix.

robinst:
Hey! I'm so sorry to hear that, I can't imagine what that must be like. Hope things are looking better now.

Issue #258, opened 2022-03-16
 Customize field conversion


JohnJyong:
**Is your feature request related to a problem? Please describe.**
I want to set some markdown fields not to be translated into html
ex：if I exclude the h1 label， ``` # test ```  will not be translated to ```<h1>test<h1>```  but  ``` <p># test<p>``` 

**Describe the solution you'd like**
we can add a field like 'includedLabelList' in the class HtmlRenderer, check the label list when doing the html rendering.

robinst:
Hm, have you tried using `enabledBlockTypes` on the `Parser.Builder`?: https://www.javadoc.io/static/org.commonmark/commonmark/0.18.2/org/commonmark/parser/Parser.Builder.html#enabledBlockTypes-java.util.Set-

Try supplying every block type except `Heading.class`.

JohnJyong:
I have tried the `enabledBlockTypes ` ,  However, it can't pinpoint a specific block，such as `Heading `
Includes `h1 h2 h3`... `ListBlock` includes `ol ul`

JohnJyong:
> Hm, have you tried using `enabledBlockTypes` on the `Parser.Builder`?: https://www.javadoc.io/static/org.commonmark/commonmark/0.18.2/org/commonmark/parser/Parser.Builder.html#enabledBlockTypes-java.util.Set-
> 
> Try supplying every block type except `Heading.class`.

I have tried the `enabledBlockTypes ` ,  However, it can't pinpoint a specific block，such as `Heading `
Includes `h1 h2 h3`... `ListBlock` includes `ol ul`. 
 When I tried to pass the subset of them , the code below 
```
        Parser parser = Parser.builder()
                .extensions(extensions)
                .enabledBlockTypes(new HashSet<>(Arrays.asList(BlockQuote.class, OrderedList.class)))
                .build();
```

it will cause the nullPoint exception below
```
Exception in thread "main" java.lang.NullPointerException
	at org.commonmark.internal.DocumentParser.findBlockStart(DocumentParser.java:438)
	at org.commonmark.internal.DocumentParser.parseLine(DocumentParser.java:230)
	at org.commonmark.internal.DocumentParser.parse(DocumentParser.java:109)
	at org.commonmark.parser.Parser.parse(Parser.java:70)
```

Is my usage wrong? pls help , thanks in advance @robinst

JohnJyong:
[...]

PR #262, opened 2022-06-02
Check arg of enabledBlockTypes when building parser instead of NPE later

robinst:
Came up in #258, see https://github.com/commonmark/commonmark-java/issues/258#issuecomment-1143046809


boyshell:
您好！邮件已收到。

Issue #277, opened 2022-11-16
Support for <video> tags


bplump:
We'd like to embed videos in our CommonMark content. We currently have the `![name](link)` syntax to generate an `<img>` tag. 

[This CommonMark extension discussion/proposal](https://talk.commonmark.org/t/embedded-audio-and-video/441) doesn't seem to have gone anywhere, but it seems to coalesce around using the filename to auto-detect whether a `<video>` or `<audio>` tag should be generated over an `<img>`.

Alternatively, I was considering using image attributes to specify if the renderer should generate a `<video>` tag.

It appears that LeanPub's Markua uses [a combination of the two](https://leanpub.com/markua/read#video): auto-detecting video formats but also allowing users to override the detected type with an attribute.

I'll try to put a PR forward to kick off a discussion about implementation but if you've got early thoughts, I'm all ears!

robinst:
Hey! We already have an image attributes extension, but it seems to work in a different way to Markua's (where the attributes come before the image in a block). Not sure about extending that if we'd be the only ones with that syntax.

I was gonna say "What does GitHub do?" but it seems like they auto-detect video URLs if the video is hosted by them, but maybe that would also be an option for you? You could do that already by post-processing `Link` nodes.

Auto-detecting based on the filename should already be possible with the current API by overriding the rendering of `Image` nodes (or replacing them with a custom node in a post processor and then rendering that). Not super nice but workable.

Let me know what you're thinking of doing? Or do a PR if that's easier :)

bplump:
We opted to hack our own solution in the frontend... (eww) so I'll close this.

PR #278, opened 2022-11-17
Add support for auto-detecting video content (#277)

bplump:
Here's a first cut at auto-detecting video content. Let me know what you think of this approach.

If the attached file has an `mp4` or `webm` extension, we'll generate a `<video>` tag instead of an `img` tag. The tests are probably the best demonstration of that.

Right now, this is not configurable or overridable. I will try to use image attributes so that `{type=image}` or `{type=video}` forces the tag type but my best solution so far feels pretty hacky... (`AttributeProvider.setAttributes` wants to extend attributes and know the tag name... but I don't have the attributes to extend or know the tag name because I need `type` to know that - Ignoring those problems, the solution works, it just feels dirty)

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
@@ -10,6 +10,9 @@ import java.util.*;
  */
 public class CoreHtmlNodeRenderer extends AbstractVisitor implements NodeRenderer {
 
+    private static final Set<String> VIDEO_EXTENSIONS = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList("mp4", "webm")));
+
     protected final HtmlNodeRendererContext context;
     private final HtmlWriter html;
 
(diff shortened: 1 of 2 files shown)


commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
```diff
@@ -274,6 +276,44 @@ private void renderListBlock(ListBlock listBlock, String tagName, Map<String, St
         html.line();
     }
 
+    private static MediaType detectMediaType(String url) {
+        int extensionStart = url.lastIndexOf('.');
```
robinst, line 280: This won't work if the URL happens to have query params that include a dot, for example `/url.mp4?param=foo.bar`. We probably need to parse the URL here and get the extension of the path that way.

robinst (commented):
One inline comment, if we decide to go that way.

Have you looked at how hard it is to implement this without changes in commonmark-java? It should be possible to provide a custom renderer and take over rendering of `Image` nodes. There's an example here: https://github.com/commonmark/commonmark-java#customize-html-rendering

Having said that, if this is generally useful it might be worth including this (with an option for `HtmlRenderer`).

bplump:
> Have you looked at how hard it is to implement this without changes in commonmark-java?

I haven't tried that. Given that there doesn't seem to be any standard syntax for generating an embedded video, maybe that's the best option.

bplump:
> Have you looked at how hard it is to implement this without changes in commonmark-java?

I've not, but that's definitely worth looking into given there doesn't seem to be much consensus on how this should be done.

I'm going to park this for now. I probably won't be able to get back to it for a while and it needs some more investigation before progressing. Thanks for the feedback and help!

robinst:
Ok 👍. Yeah I wish there was a commonly accepted syntax, then it would be a no-brainer to implement that.

PR #288, opened 2023-03-25
Fix #287 - Eating of whitespace at start of line

### LINKED ISSUE #287: Consistently makes element-android formatting of code blocks wrong
Steps to reproduce the problem (provide example Markdown if applicable):

```cpp
struct Foo {
  Foo();
  int a;
}
```

Expected behavior: 
The behavior should be identical to what you see in this window. However, it consistently is not. 

```
struct Foo {
  Foo();
  int a;
}
```

Actual behavior:

```
struct Foo {
Foo();
int a;
}
```

The reason why I've found this problem is because [element-android](https://github.com/vector-im/element-android) depends on this library - and on the android app the code blocks consistently (for everyone _should_ be formatted erroneously). It's a bit strange that more people have not noticed this (to me). For Python code for instance, it makes the code completely unintelligible.

Here's what the common mark spec outputs (which is what should be expected): [https://spec.commonmark.org/dingus/](https://spec.commonmark.org/dingus/?text=%23%23%20Try%20CommonMark%0A%0A%60%60%60cpp%0Astruct%20Foo%20%7B%0A%20%20Foo()%3B%0A%20%20int%20a%3B%0A%7D%0A%60%60%60)

I think I've identified where the problem is and will be submitting a PR.

theIDinside:
In addLine() we return the substring "Foo()" when it should be "  Foo()".

This patch removes taking the substring and instead returns the line as is.

The prior if branch ([here](https://github.com/theIDinside/commonmark-java/commit/4fecb26a68a279bed6c60ab62274de8cd3fc16c4#diff-9c1169707cc7d5ce9e51b82bf59c5bef0eaa5dfd8ceb611d45334ee70391cdf5R403-R413)) seem to suggest that this _is_ the expected behavior since it converts tabs to spaces, yet does not remove those.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
@@ -411,10 +411,8 @@ public class DocumentParser implements ParserState {
             }
             sb.append(rest);
             content = sb.toString();
-        } else if (index == 0) {
-            content = line.getContent();
         } else {
-            content = line.getContent().subSequence(index, line.getContent().length());
+            content = line.getContent();
         }
         SourceSpan sourceSpan = null;
         if (includeSourceSpans == IncludeSourceSpans.BLOCKS_AND_INLINES) {


AfzalivE:
Doesn't this change fail a lot of tests?

robinst:
> Doesn't this change fail a lot of tests?

Yes it does. I'm not sure what problem it's trying to solve, I've replied on #287

PR #290, opened 2023-04-07
Keep leading spaces in paragraphs when there's a list

AfzalivE:
When there are no `blockParserFactories`, leading spaces should be preserved. Here's a use case:

### Input
```
- one
  - two
```

When this is parsed with a Parser without any `enabledBlockTypes`, before this PR, it results in this:

```
- one
- two
```

But it should preserve the list hierarchy with the leading whitespace.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/DocumentParser.java
@@ -237,7 +237,9 @@ public class DocumentParser implements ParserState {
 
             BlockStartImpl blockStart = findBlockStart(blockParser);
             if (blockStart == null) {
-                setNewIndex(nextNonSpace);
+                if (!blockParserFactories.isEmpty()) {
+                    setNewIndex(nextNonSpace);
+                }
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


AfzalivE:
Ping @robinst, would you please be able to review this? Thanks!

robinst:
Note that this doesn't really have anything to do with lists, it's the same for a paragraph like this:

```
aaa
  bbb
    ccc
```

The reason for this is in the spec, leading spaces are skipped: https://spec.commonmark.org/0.31.2/#example-222
Final spaces in a paragraph are also skipped, see here: https://spec.commonmark.org/0.31.2/#example-226

Having said that, we should probably have a way to preserve these skipped spaces in the nodes somehow. Some ideas:

- Add a `SkippedSpaces` node that HTML rendering ignores but other renderers can make use of (e.g. the new `MarkdownRenderer` might also want to render it).
- Add some properties to `Text` nodes to indicate any leading/trailing spaces. Not sure how feasible this is with how inline parsing works. Not sure if `Text` would be enough, what if the line starts/ends with an inline element (e.g. emphasis, link etc).
- Add some properties to `Paragraph` nodes to indicate any leading/trailing spaces. This would have to store the line numbers somehow, and might not be super convenient to consume.

We also need to consider interaction with e.g. `getMarkerIndent` on `ListItem`. Maybe that could be replaced by whatever we build for this if it's generic enough.

robinst:
Closing this as it's not the right approach, see comment above and https://github.com/commonmark/commonmark-java/issues/301.

PR #296, opened 2023-07-14
add column width to TableCell nodes

dwyand:
Adds an extra property for column with to `TableCell` nodes.

The width is determined by the number of dash and colon characters in the separator line of a table, and passed on to the consumer as an integer. This permits consumers of the AST to render tables that are more proportional the width in the source markdown.

No changes have been made to the HTML render to make use of the width information, but the test for this feature demonstrates a possible usage.

Should a cell be be created that is further to the right than the parsed separator columns are (i.e., when alignment would be null), then the width will be reported as 0.

### DIFF
--- a/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/TableCell.java
+++ b/commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/TableCell.java
@@ -9,6 +9,7 @@ public class TableCell extends CustomNode {
 
     private boolean header;
     private Alignment alignment;
+    private int width;
 
     /**
      * @return whether the cell is a header or not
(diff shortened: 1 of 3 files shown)


dwyand:
While implementing a workaround in case this does not get merged, I've realized that the widths make more sense as an addition to `TableBlock` instead of repeating them on each `TableCell`. Will update implementation, marking as draft until then.

commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/TableCell.java
```diff
@@ -32,6 +33,17 @@ public void setAlignment(Alignment alignment) {
         this.alignment = alignment;
     }
 
+    /**
+     * @return the cell width
```
robinst, line 37: Can you add a description of what this mean here, similar to the PR description?

robinst (commented):
Just one comment, otherwise looks good to merge!

commonmark-ext-gfm-tables/src/main/java/org/commonmark/ext/gfm/tables/TableCell.java
robinst: I've done that now..

robinst:
Released in 0.22.0 now: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0220---2024-03-15

Issue #125, opened 2018-03-24
Add jlink-compatible Java9/Jigsaw module-info (not automatic modules)


xzel23:
Latest version has automatic modules and can be used in Java 9 projects (#114). However, this is not compatible with the jlink tool.

__Steps to reproduce the problem (provide example input):__

- create a Java 9 based project that uses modules
- Use jlink

__Expected behavior:__

jlink links the prokect

__Actual behavior:__

jlink reports an error:
```
> Task :link FAILED
Error: automatic module cannot be used with jlink: org.commonmark.ext.gfm.strikethrough from file:///Users/axel/.gradle/caches/modules-2/files-2.1/com.atlassian.commonmark/commonmark-ext-gfm-strikethrough/0.11.0/2b3f085711511202c380f2540760d64302c6bb7e/commonmark-ext-gfm-strikethrough-0.11.0.jar

```

xzel23:
I think module-info should be added in a Java 8 compatible way. I have done it for some libraries that use the gradle build tool, so it should be possible in Maven too. If noone is working on this, I will try to do this myself, but. since I have no experience in Maven, it might take some time.

robinst:
Apparently adding a module-info breaks Android usage, quoted from [Stephen's blog post](http://blog.joda.org/2018/03/jpms-negative-benefits.html?m=1):

> And just for kicks, your project can no longer be used by Android (as the team there seems to be very slow in adding a simple "ignore module-info" rule).

robinst:
I'm not a fan of having a separate branch that needs to be kept up to date. Releasing a duplicate set of artifacts would be possible for Java 9+ (with a different group ID or artifact IDs), but that is also a lot of effort and would add confusion.

How would you feel about having a script that can process a set of jars to add the module-info? Would that work for you use case?

xzel23:
I think that would not really help the situation. IMHO JPMS has huge benefits once it's there, but we will never get there unless libraries are updated to include module-info. And that can already be done in a Java 8 compatible way. Thats's what should be done and will be of benefit for us all.

I can however see your point about Android and tools not being ready. IMHO those tools should be fixed as soon as possible. I have already started adding module.info in my own fork of commonmark and maybe will even publish jars (with version "xxx-JPMS" on bintray) that are still compatible with Java 8, but not android and the tools you mentioned.

Once Java 11 (the next LTS) is out, we should look at this again and the issues with android and tools are solved, can merge my changes into master.

[...]

PR #299, opened 2023-10-02
Raise minimum version to Java 11 (and test on 21 as well)

robinst:
Bump `actions/setup-java` to `v2` as well, which requires specifying a distribution and use 8 instead of 1.8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,15 +9,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [1.8, 11, 17]
+        java: [8, 11, 17, 21]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
 
       - name: Set up JDK
-        uses: actions/setup-java@v1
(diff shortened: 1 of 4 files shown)


robinst:
Ok, will have to raise the minimum Java version first, see error on 21:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project commonmark-test-util: Compilation failure: Compilation failure: 
Error:  Source option 7 is no longer supported. Use 8 or later.
Error:  Target option 7 is no longer supported. Use 8 or later.
```

robinst:
Let's go to Java 11, see https://github.com/commonmark/commonmark-java/issues/125#issuecomment-1936780375

Reasons:
* Java 8 was released almost 10 years ago
* We need 9+ to build with modules (unless we use a compatibility plugin like ModiTect; but I don't want that as it complicates the build and moves module-info to poms instead of normal files with IDE support, etc)
* We could go to 9 but it doesn't seem like many people use 9 and not 11. 11 is also a LTS version so is better supported.

Issue #12, opened 2015-08-10
CommonMark/Markdown renderer (round-trip parsing/rendering)


robinst:
It's useful to be able to render a parsed document back to CommonMark markdown. This is a tracker issue for issues relating to that, so far:
- [x] Unable to disambiguate emphasis delimiter #10
- [x] Unable to disambiguate list-item type #11

pcj:
More currently unanswerable questions once the parse tree has been constructed:

- [ ] Headers: was it ATX or Setext type?
- [ ] Links: was it an inline, reference, or auto-link?

There are other metadata items for different types of nodes that would need to be captured to re-render a completely identical copy of the input text, things like:

- [ ] The opener and closer sequences for certain items like headers, inline code, fenced code blocks (as these can be variable to some degree).
- [ ] Whitespace padding between block elements (would be addressed by #1).

Finally, reference link definitions are swallowed up completely by the ParagraphParser, no AST nodes exist for them.  Another strategy would be to leave those in somehow and put them in the visitor such that renderers have the option to deal with them.  I'm guessing that this is not an easy issue to address since link reference defintions can exist in many block contexts (paragraph, list-item, blockquote), forcing all those to have to parse them.

- [ ] Reference link definitions: define an AST node type.

PaulKlint:
In the Rascal project (https://github.com/usethesource/rascal) I am using your very nice framework and have also felt the need for a round-trip back to CommonMark. Sofar, I have written an `AnsiRenderer` that maps back to text (and optionally to Ansi characters for highlighting). This is acceptable with a somewhat relaxed definition of round-trip but it breaks down when trying to apply this to the `Tables` extension. There are several issues in that case:
- [ ] The general problem that all `*Extension` classes also need to implement `AnsiRendererExtension` (or whatever the round-trip renderer will be named)
- [ ] More specifically, the meta-data to reconstruct the table header is problematically missing and cannot be easily approximated.

robinst:
@PaulKlint: Cool!

Yeah, if you have external custom renderers for extensions, you either need to depend on the artifacts or provide additional artifacts for each extension. If it's a built-in renderer, we can just implement them for the extensions directly.

What specifically is missing from the metadata for table headers? Could you create a separate issue for that?

mattsheppard:
Is there in fact an implementation of a CommonMark renderer (with whatever limitations there are as mentioned above), or is the intention only to develop one once they are overcome?

[...]

PR #306, opened 2024-02-09
Add renderer producing Markdown (CommonMark)

robinst:
Adds a `MarkdownRenderer` that renders a `Node` tree to CommonMark Markdown. Supports all core nodes as well as "gfm-strikethrough", "gfm-tables" and "ins" extensions. This is very useful for applications that want to produce Markdown, e.g.:

* Parse some input Markdown, modify the nodes (AST), then render back to Markdown
* Create a `Node` tree via code and then render that to Markdown

The goal is that `Node tree -> render to Markdown -> parse to Node tree` results in an equivalent Node tree. That is verified with all the examples from the spec and some manual test cases.

However it does not mean that a `Markdown -> parse to Node tree -> render to Markdown` round-trip produces the same Markdown as the original. The parse step does not preserve enough detail for that yet, e.g. things like insignificant whitespace or what characters were escaped.

___

In the future, we can make changes that preserve more details when parsing and make use of that to bring the rendered Markdown closer to the original. But the Markdown renderer can't rely on those extra details to work correctly, as that would break the use case where there is no parsing (when creating `Node` tree via code).

### DIFF
--- a/commonmark-ext-gfm-strikethrough/src/main/java/org/commonmark/ext/gfm/strikethrough/Strikethrough.java
+++ b/commonmark-ext-gfm-strikethrough/src/main/java/org/commonmark/ext/gfm/strikethrough/Strikethrough.java
@@ -4,19 +4,23 @@ import org.commonmark.node.CustomNode;
 import org.commonmark.node.Delimited;
 
 /**
- * A strikethrough node containing text and other inline nodes nodes as children.
+ * A strikethrough node containing text and other inline nodes as children.
  */
 public class Strikethrough extends CustomNode implements Delimited {
 
-    private static final String DELIMITER = "~~";
+    private String delimiter;
+
+    public Strikethrough(String delimiter) {
+        this.delimiter = delimiter;
+    }
 
     @Override
     public String getOpeningDelimiter() {
-        return DELIMITER;
+        return delimiter;
     }
 
     @Override
     public String getClosingDelimiter() {
-        return DELIMITER;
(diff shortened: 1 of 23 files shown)


Apple1D:
> Adds a `MarkdownRenderer` that renders a `Node` tree to CommonMark Markdown. Supports all core nodes as well as "gfm-strikethrough", "gfm-tables" and "ins" extensions. This is very useful for applications that want to produce Markdown, e.g.:
> 
> * Parse some input Markdown, modify the nodes (AST), then render back to Markdown
> * Create a `Node` tree via code and then render that to Markdown
> 
> The goal is that `Node tree -> render to Markdown -> parse to Node tree` results in an equivalent Node tree. That is verified with all the examples from the spec and some manual test cases.
> 
> However it does not mean that a `Markdown -> parse to Node tree -> render to Markdown` round-trip produces the same Markdown as the original. The parse step does not preserve enough detail for that yet, e.g. things like insignificant whitespace or what characters were escaped.
> 
> See #12.
> 
> ___
> 
> In the future, we can make changes that preserve more details when parsing and make use of that to bring the rendered Markdown closer to the original. But the Markdown renderer can't rely on those extra details to work correctly, as that would break the use case where there is no parsing (when creating `Node` tree via code).

robinst:
Released in 0.22.0 now: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0220---2024-03-15

Issue #263, opened 2022-06-22
Add support for inline subparsers


avarga:
I was trying to implement an extension that turns references denoted by a prefix character (@,#, or ~) into a link, and missing the means to do it. A PostProcessor won't do it, because I want the prefix characters to be escapable by backslash to take away their special meaning, and the info necessary for that doesn't survive into the AST. The Delimiter stuff isn't useful either, because we have no definite closing character.

The proper solution would be a sub-parser that is activated then the main parser sees the prefix character. This something that `InlineParserImpl` already does internally:

```
        this.inlineParsers = new HashMap<>();
        this.inlineParsers.put('\\', Collections.<InlineContentParser>singletonList(new BackslashInlineParser()));
        this.inlineParsers.put('`', Collections.<InlineContentParser>singletonList(new BackticksInlineParser()));
        this.inlineParsers.put('&', Collections.<InlineContentParser>singletonList(new EntityInlineParser()));
        this.inlineParsers.put('<', Arrays.asList(new AutolinkInlineParser(), new HtmlInlineParser()));
  ```

but not only is `InlineParserImpl` in an `internal` package, but all relevant methods and fields are private as well, so I have no chance to use it even if I didn't care about backward compatibility.

I suggest opening up an official way of adding sub-parsers to the internal parser.

robinst:
Yeah it would be useful, but I've been hesitant to commit to an API for it before (in order not to have to break API if we need to change it). One major outstanding change is to enable round-trip parsing which will probably affect the API.

But we could enable support for it while keeping it "internal", meaning "It's not API yet and if you use it you might have to adjust your code on newer versions". I won't have time to work on it in the next few weeks though.

robinst:
I just merged a change to implement this, see https://github.com/commonmark/commonmark-java/pull/321

I'll keep this open until the change is released.

robinst:
🎉  Released in 0.23.0 now, see `customInlineContentParserFactory` in `Parser.Builder`. Full changelog: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0230---2024-09-16

PR #319, opened 2024-04-10
Allow to add inline parsers to built-in implementation

### LINKED ISSUE #317: Question: How to extract original source between delimiters in a delimiterProcessor?
Hello, thanks for this well written library! We're considering it for replacing the parser in our [clojure _markdown document to data_ library](https://github.com/nextjournal/markdown).

I'm trying to add support for inline latex formulas delimited between single dollar signs `some $\phi$ ormula` via a custom delimiterProcessor but couldn't find examples of such a processor which is able to extract the original source between the delimiters since afaiu delimiter processors operate _after_ the inline parsing.  

What is currently possible is to capture and join all literals from text nodes between the opener and the closer `Nodes.between()` but that might not reproduce the exact source as in 
```
$some [stuff](/which/might/looklike/a/link) say$
```
same using a Markdown or Text renderer I guess.

I guess this falls under #263…

zampino:
This is a minimal proposal for allowing the current internal implementation to customize inline parsing of special characters, following @robinst advice to provide a starting POC :-).

I quickly tested this in combination with `Parser.Builder.inlineParserFactory` in the context of a clojure application for parsing inline latex delimited between pairs of $ characters, but proper testing is still missing. It's a first step to see what's possible, far from providing enough fine-grained customization.

Addresses #263. Closes #317.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
+++ b/commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
@@ -76,6 +76,12 @@ public class InlineParserImpl implements InlineParser, InlineParserState {
         return scanner;
     }
 
+    public InlineParser addInlineParser(Character c, List<InlineContentParser> parsers) {
+        this.inlineParsers.put(c, parsers);
+        this.specialCharacters.set(c);
+        return this;
+    }
+
     private static void addDelimiterProcessors(Iterable<DelimiterProcessor> delimiterProcessors, Map<Character, DelimiterProcessor> map) {
         for (DelimiterProcessor delimiterProcessor : delimiterProcessors) {
             char opening = delimiterProcessor.getOpeningCharacter();


commonmark/src/main/java/org/commonmark/internal/InlineParserImpl.java
```diff
@@ -76,6 +76,12 @@ public Scanner scanner() {
         return scanner;
     }
 
+    public InlineParser addInlineParser(Character c, List<InlineContentParser> parsers) {
+        this.inlineParsers.put(c, parsers);
```
robinst, line 80: So this will currently override any existing built-in parsers with that same special character. I think we should do the same thing as for block parsers, see https://github.com/commonmark/commonmark-java/blob/1bf85c6c240bb9a5ab16a931595e83db5526c4e2/commonmark/src/main/java/org/commonmark/parser/Parser.java#L208

So try the custom parsers first. The custom parser can decide not to handle something, in which case the built-in parsers are tried next.

robinst (changes requested):
So in order for users to actually use this, the configuration needs to be on the `Parser.Builder`, similar to custom block parsers. Then when constructing the inline parser, pass it in the constructor.

(We will also need to move some classes into an exported package, but can do that later.)

robinst:
Closing this in favor of
- https://github.com/commonmark/commonmark-java/pull/321

PR #328, opened 2024-05-14
README: Bump version

kmh916:
update version in README.md `0.21.0` to `0.22.0`


sebthom:
@robinst would you mind merging this? I just felt into the same trap of depending on the wrong version and then wondering why `org.commonmark.renderer.markdown.MarkdownRenderer` from the example cannot be found on classpath.

robinst:
Merged, thanks!

PR #332, opened 2024-07-07
Footnotes extension

### LINKED ISSUE #273: Support for footnotes
[Footnotes](https://www.markdownguide.org/extended-syntax/#footnotes) are arguably an important markup feature but doesn't appear to be supported by CommonMark yet. GFM [supports](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#footnotes) them and we're seeing an increasing number of uses e.g. in [Chromium](https://source.chromium.org/search?q=%22%5B%5E%22%20file:.md&ss=chromium) and [Android](https://cs.android.com/search?q=%22%5B%5E%22%20file:.md&sq=) but these don't render correctly in [Google Code Search](https://source.chromium.org/chromium/chromium/src/+/main:third_party/tflite/src/tensorflow/core/ir/README.md) nor [gitiles/Gerrit](https://swiftshader.googlesource.com/SwiftShader/+/refs/changes/49/69249/2/README.md).

(Google bug b/255316523, contact hanwen@ for integration when fixed)

robinst:
This adds a new extension `commonmark-ext-footnotes` (class `org.commonmark.ext.footnotes.FootnotesExtension`) to implement footnotes syntax as in GitHub Flavored Markdown (see [docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes)). Fixes #273.

An example:

```
Some text with a footnote[^1].

[^1]: The text of the footnote.
```

The `[^1]` is parsed as a `FootnoteReference`, with `1` being the label.
The line with `[^1]: ...` is a `FootnoteDefinition`, with the contents as child nodes (can be a paragraph like in the example, or other blocks like lists).

Apart from the parsing, the extension also comes with rendering of footnotes for HTML and Markdown.

## Extension mechanisms

In order to implement this as a separate extension, the following APIs were added to commonmark core:
- `DefinitionMap`: New class for storing and looking up definitions by a label, with label normalization as for link reference definitions
- `BlockParser`: New method `getDefinitions` that can be implemented to return definitions that can later be accessed during inline parsing (the built-in `ParagraphParser` also uses that mechanism now; previously it was a special case in the parser)
- `LinkProcessor`: New interface that can be implemented to customize link/image processing. This is used to turn `[^1]` into `FootnoteReference` nodes.
- `NodeRenderer`: New methods `beforeRoot` and `afterRoot` that are called before/after rendering a document; used to render footnotes at the end of the document

### PostProcessor

Could footnote reference parsing have been implemented as a `PostProcessor` step after inline parsing? No, because a foonote reference like `[^*foo*]` would have been turned into emphasis by inline parsing, whereas footnote parsing needs the raw `*foo*` as a label.

### InlineContentParser

I considered using the recently-added inline parsing customization API, using `[` as the trigger character. That would work for simple cases, but not for others. E.g. in this:

```
[^foo](/url)

[^foo]: note
```

That is not a footnote followed by `(/url)`, but instead it's an inline link. In other words, if parsing as a link is possible, that is preferred.

That means our custom inline parser for `[` would have to be able to parse the full link syntax in order to give preference to links, which is quite tricky. In addition to that, it would have have to trigger on `!`, for a footnote like `![^foo]`, which normally would be parsed as an image node.

So that's what `LinkProcessor` solves: It keeps the tricky link parsing in the inline parser, but allows extensions to decide to treat certain things not as links, but different types of nodes, or maybe even parse things that come after a link (e.g. image attributes could be implemented on top of this).


robinst:
I've found some interesting behaviors (bugs) in GitHub's implementation while working on this. E.g.:

```
[^1]: One [^2]
[^2]: Two
```

That shouldn't render anything, because only 2 is referenced but from another footnote that is not rendered. But GitHub renders "Two" as the only footnote, with a "back" link pointing nowhere.

Another one:

```
[^1]: One [^2]

Test [^1]

[^2]: Two
```

The order of footnotes should be One (referenced first in the text), then Two (referenced from a footnote). But GitHub renders Two first (because it finds the `[^2]` reference first).

---

I've decided not to follow GitHub's implementation for these edge cases, but instead go for the nicest result. See docs on [FootnoteHtmlNodeRenderer.java](https://github.com/commonmark/commonmark-java/pull/332/files#diff-f7fa4d2cbbed7de7c83fb541a0a5c3a1c3a0388ff0e33f88ccc528ded9746e7b) for more about this. If bug-for-bug-compatibility is required at some point it should be simple enough to add as an option.

PR #333, opened 2024-07-08
Render ThematicBreak literal in markdown

### LINKED ISSUE #331: ThematicBreak literal is lost in the Markdown renderer
Corresponding to [the specification](https://spec.commonmark.org/0.30/#thematic-breaks), thematic breaks can consist of 3 or more consecutive on of characters `-`, `_` or `*` with 0-3 leading spaces (i.e. regex `^\s{0,3}[-_*]{3,}$`).
This logic seem to be correct when reading Markdown - the thematic breaks are correctly recognised and captured as `ThematicBreak` node with `literal` parameter containing the actual value from the Markdown.
However, [during rendering](https://github.com/commonmark/commonmark-java/blob/114956af772cb82aa8f5c262e735aadcb0adc9cf/commonmark/src/main/java/org/commonmark/renderer/markdown/CoreMarkdownNodeRenderer.java#L93) the `literal` is dropped and replaced with static `___`

Expected behaviour: this markdown looks the same after parse/render
```md
foo

   *******

bar

---

baz
```

Actual behaviour: rendered transforms it into
```md
foo

___

bar

___

baz
```

jumale:
Closes #331

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/markdown/CoreMarkdownNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/markdown/CoreMarkdownNodeRenderer.java
@@ -87,8 +87,12 @@ public class CoreMarkdownNodeRenderer extends AbstractVisitor implements NodeRen
 
     @Override
     public void visit(ThematicBreak thematicBreak) {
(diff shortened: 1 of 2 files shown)


robinst:
Thanks for the fix, will be included in the next release :).

PR #358, opened 2025-01-05
fix: Add 'requires transitive' to fix compiler warning

sebthom:
This fixes the compiler warning "The type Extension from module org.commonmark may not be accessible to clients due to missing 'requires transitive'"

### DIFF
--- a/commonmark-ext-autolink/src/main/java/module-info.java
+++ b/commonmark-ext-autolink/src/main/java/module-info.java
@@ -1,6 +1,6 @@
 module org.commonmark.ext.autolink {
     exports org.commonmark.ext.autolink;
 
-    requires org.commonmark;
+    requires transitive org.commonmark;
     requires org.nibor.autolink;
 }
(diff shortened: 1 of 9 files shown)


robinst:
Where does the compiler warning show up?

(If a consumer depends on one of the extensions, they will also need to depend on the core. Using just the extension without any core types doesn't make any sense.)

sebthom:
I am seeing this in Eclipse. The reason is that the extensions leak/expose classes from their transitive dependency (commonmark).

For example:
```java
public class AutolinkExtension implements Parser.ParserExtension {

    private AutolinkExtension() {
    }

    public static Extension create() {
        return new AutolinkExtension();
    }

    @Override
    public void extend(Parser.Builder parserBuilder) {
        parserBuilder.postProcessor(new AutolinkPostProcessor());
    }

}
```
When I do `AutolinkExtension.create()` I get an instance of type `Extension`. But `Extension` is not part of the jar providing class `AutolinkExtension` it is a transitive dependency of that jar thus the `transitive` keyword is required.

robinst:
Right, I see. Maybe it was a mistake for `create` to return `Extension` (instead of the more specific type `AutolinkExtension`). But anyway, no reason to change that now. Will go ahead and merge the PR.

PR #361, opened 2025-01-30
Enable adding own markdown renderers in builder

kabir:
The NodeRendererMap refuses to write over the existing entries.

Looking a bit into it, it was because MarkdownRenderer.RendererContext iterates over the nodeRenderFactories in reverse order. This means that the last (i.e. the default) one is added first, making it impossible to override.

There were a few ways to fix this, but looking at the corresponding code in TextContentReader.RenderContext and HtmlRenderer.RenderContext, they iterate over the list of factories in the natural order.

@robinst @sebthom if you agree with the fix, it would be great if it could be included in a release soon :-)


robinst:
Thanks @kabir! I wanted some changes around the tests, so raised #368 to fix this instead (the actual fix is the same though).

robinst:
@kabir FYI the fix has been released in 0.25.0 now: https://github.com/commonmark/commonmark-java/releases/tag/commonmark-parent-0.25.0

PR #364, opened 2025-02-13
generic Renderer for render targets richer than String

rpm0372:
Changed `org.commonmark.node.Node.Renderer` to `Renderer<T>` and updated all three type descendants to concretize as `Render<String>`.

This change was motivated by my project's need to render markdown to targets not easily represented as a single `String`.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/Renderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/Renderer.java
@@ -2,7 +2,7 @@ package org.commonmark.renderer;
 
 import org.commonmark.node.Node;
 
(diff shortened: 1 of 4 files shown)


robinst:
> This change was motivated by my project's need to render markdown to targets not easily represented as a single `String`.

Can you explain this more? So you want to be able to have a `Renderer<Something>`? Why can't you define your own interface (and delegate to commonmark's `Renderer`)?

rpm0372:
Because commonmark's Renderer interface only supports returning a String.

robinst:
Why is that a problem? What are you trying to solve?

PR #365, opened 2025-02-20
Change ordering of `node.getNext()` and `context.render(node)` in CoreHtmlRenderer.java

silentbaws:
Getting the next node before rendering the current one prevents a renderer from contextually unlinking the next node.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
+++ b/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java
@@ -252,9 +252,8 @@ public class CoreHtmlNodeRenderer extends AbstractVisitor implements NodeRendere
     protected void visitChildren(Node parent) {
         Node node = parent.getFirstChild();
         while (node != null) {
-            Node next = node.getNext();
             context.render(node);
-            node = next;
+            node = node.getNext();
         }
     }
 


robinst:
Hmm, not sure how I feel about unlinking nodes during rendering. Can you explain how you're marking use of this? I'm wondering if it would be better to do this in a preprocessor step (e.g. could replace both nodes with a custom one).

silentbaws:
> Hmm, not sure how I feel about unlinking nodes during rendering. Can you explain how you're marking use of this? I'm wondering if it would be better to do this in a preprocessor step (e.g. could replace both nodes with a custom one).

On mobile so can’t do a full write up of what I’m running in to but the gist is. I’m trying to render obsidian links/images.

I have a custom link processor for that [here](https://github.com/silentbaws/website/blob/mainline/src/main/java/com/davisellwood/website/common/components/markdown/parsing/CustomLinkProcessor.java#L19) but this has a side effect of creating nodes that I have to collapse hackily during rendering [here](https://github.com/silentbaws/website/blob/mainline/src/main/java/com/davisellwood/website/common/components/markdown/parsing/CustomLinkProcessor.java#L20).

robinst:
Hey @silentbaws. Both places you linked are just in `CustomLinkProcessor`, did you mean to link somewhere else? Do you have an example of what syntax you're trying to parse and how you want to render it? (I had a look at your repo but couldn't see any tests.)

PR #406, opened 2025-10-04
Allow Detection of WWW Autolinks

rdestefa:
Would be nice to have this additional functionality, so took a stab at a PR for this.

### DIFF
--- a/commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/AutolinkExtension.java
+++ b/commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/AutolinkExtension.java
@@ -1,5 +1,8 @@
 package org.commonmark.ext.autolink;
 
+import java.util.EnumSet;
+import java.util.Set;
+
 import org.commonmark.Extension;
 import org.commonmark.ext.autolink.internal.AutolinkPostProcessor;
 import org.commonmark.parser.Parser;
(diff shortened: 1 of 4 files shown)


commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/AutolinkExtension.java
```diff
@@ -18,16 +21,75 @@
...
+         * @return {@code this}
+         */
+        public Builder linkTypes(AutolinkType... linkTypes) {
+            if (linkTypes == null) {
+                throw new NullPointerException("linkTypes must not be null");
+            }
+
+            if (linkTypes.length == 0) {
+                throw new IllegalArgumentException("linkTypes must not be empty");
+            }
```
robinst, line 65: No need to duplicate the checks above if we're delegating to `linkTypes(Set<AutolinkType> linkTypes)` anyway.

commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/internal/AutolinkPostProcessor.java
```diff
@@ -11,9 +12,40 @@
...
 public class AutolinkPostProcessor implements PostProcessor {
 
-    private LinkExtractor linkExtractor = LinkExtractor.builder()
-            .linkTypes(EnumSet.of(LinkType.URL, LinkType.EMAIL))
-            .build();
+    private final LinkExtractor linkExtractor;
+
+    public AutolinkPostProcessor() {
+        this(EnumSet.of(AutolinkType.URL, AutolinkType.EMAIL));
+    }
```
robinst, line 19: This constructor shouldn't be necessary, as this is an internal class (in `internal` package).

commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/internal/AutolinkPostProcessor.java
```diff
@@ -11,9 +12,40 @@
...
+    public AutolinkPostProcessor(Set<AutolinkType> linkTypes) {
+        if (linkTypes == null) {
+            throw new NullPointerException("linkTypes must not be null");
+        }
+
+        if (linkTypes.isEmpty()) {
+            throw new IllegalArgumentException("linkTypes must not be empty");
+        }
+
+        EnumSet<LinkType> types = EnumSet.noneOf(LinkType.class);
```
robinst, line 30: We can use `var` nowadays:

```suggestion
        var types = EnumSet.noneOf(LinkType.class);
```

commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/internal/AutolinkPostProcessor.java
```diff
@@ -67,8 +99,12 @@ private static Text createTextNode(String literal, Span span, SourceSpan sourceS
     }
 
     private static String getDestination(LinkSpan linkSpan, String linkText) {
-        if (linkSpan.getType() == LinkType.EMAIL) {
+        LinkType type = linkSpan.getType();
```
robinst, line 102:
```suggestion
        var type = linkSpan.getType();
```

commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/internal/AutolinkPostProcessor.java
```diff
@@ -67,8 +99,12 @@ private static Text createTextNode(String literal, Span span, SourceSpan sourceS
...
+
+        if (type == LinkType.EMAIL) {
             return "mailto:" + linkText;
+        } else if (type == LinkType.WWW) {
+            return "http://" + linkText;
```
robinst, line 107: Hmm, why `http://` and not `https://`? I feel like nowadays we should default to the latter. If anyone wants `http://`, we could add it as an another option to the extension builder.

commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/AutolinkExtension.java
rdestefa: Removed the second check. The first check was just to provide a more explicit error message than whatever is thrown by calling `Set.of(null)`, but I can remove if we think that's unnecessary.

commonmark-ext-autolink/src/main/java/org/commonmark/ext/autolink/internal/AutolinkPostProcessor.java
robinst: Hmm ok, let's copy them in that case.

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
 
+## Unreleased
+### Added
+- Autolink extension: Now supports configuration of different link types that
+  should be recognized and converted to links. See `AutolinkExtension#builder`
+  | Type    | Default? | Description                                               |
+  |---------|----------|-----------------------------------------------------------|
+  | `URL`   | Yes      | URL with a protocol such as `https://example.com`         |
+  | `EMAIL` | Yes      | Email address such as `foo@example.com`                   |
+  | `WWW`   | No       | An address beginning with `www` such as `www.example.com` |
```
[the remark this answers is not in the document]
robinst: Interesting. Yeah ok let's enable it by default. Make sure to highlight this, and add a code snippet for how to get the old behavior back.

PR #420, opened 2026-03-03
Add GFM alerts extension

### LINKED ISSUE #327: GFM admonition blocks
**Is your feature request related to a problem? Please describe.**

Github now support admonition blocks

```markdown
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

**Describe the solution you'd like**
A clear and concise description of what you want to happen. Include source code if possible.

Ideally adding a new gfm extension to comon-mark makes sense.

**Describe alternatives you've considered**

No alternative.

**Additional context**
Add any other context about the feature request here.

Source [[1]](https://github.com/orgs/community/discussions/16925#discussion-4085374), [[2]](https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/)

ia3andy:
## Note
I have used Claude to code most of it, I am currently reviewing the generated code which seems to be overall valid. I've requested to follow the pattern of existing extensions (and compare to other implementations of this feature).

## Summary
- Adds new extension module for GitHub Flavored Markdown alerts
- Implements alert detection as a `PostProcessor` (transforms `BlockQuote` nodes into `Alert` nodes after parsing)
- Includes HTML and Markdown renderers
- Supports all five standard GFM alert types: NOTE, TIP, IMPORTANT, WARNING, CAUTION
- Case-insensitive type matching (`[!note]`, `[!Note]` work like `[!NOTE]`)
- Supports custom alert types and title overrides via builder API (e.g., for localization)
- Empty/whitespace-only alerts render as normal blockquotes (matching GitHub behavior)
- Alerts only at document level (not inside list items or nested blockquotes, matching GitHub)

## Test approach
- `AlertsSpecTest` — parameterized spec-based tests from `alerts-spec.txt`, expectations verified against GitHub Markdown API (`gh api markdown -f mode=gfm`)
- `AlertsTest` — custom types, builder validation, AST assertions
- `AlertsMarkdownRendererTest` — roundtrip rendering tests

## Changes from review
- Rewrote from custom `BlockParser` to `PostProcessor` approach
- Added case-insensitive type matching
- Allow overriding standard type titles
- Removed unnecessary `TextContentRenderer` (core handles it)
- Constructor injection for `Alert` (immutable type, no public setter)
- Moved `STANDARD_TYPES` from `Alert` node to `AlertsExtension`
- Removed `getCustomTypes()` config leak from public API
- Fixed visitor traversal (early return after successful conversion)
- Removed dead code (`capitalize` fallback)

Closes #327


robinst (changes requested):
Some first comments, mostly around parsing and edge cases.

I was curious how GitHub implements this in [cmark-gfm](https://github.com/github/cmark-gfm), and the answer seems to be that they don't:

https://github.com/github/cmark-gfm/issues/350#issuecomment-1747789268

> Alerts are not implemented in this project. They don't relate to their parser. They don't relate to syntax. So they're not here.
> 
> They are implemented as a post processing step on a sort of "DOM" version of the resulting document.

Can you explore what the implementation would look like as a `PostProcessor` instead? That has the advantage of not having to copy all the parsing logic from block quotes. (From the edge cases with nesting I commented about, it seems like they only look at top-level nodes.)

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertPostProcessor.java
```diff
@@ -0,0 +1,170 @@
...
+
+        AlertVisitor(Set<String> allowedTypes) {
+            this.allowedTypes = allowedTypes;
+        }
+
+        @Override
+        public void visit(BlockQuote blockQuote) {
+            // Only convert top-level block quotes (direct children of Document).
+            // This matches GitHub's behavior where alerts are only detected at the document level.
+            if (blockQuote.getParent() instanceof Document) {
```
robinst: If you do it this way, you're still visiting the full document tree, which is unnecessary.

Just manually iterate over the direct children of `Document` instead of using an `AbstractVisitor` please.

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
```diff
@@ -0,0 +1,117 @@
...
+         * @return {@code this}
+         */
+        public Builder addCustomType(String type, String title) {
+            if (type == null || type.isEmpty()) {
+                throw new IllegalArgumentException("Type must not be null or empty");
+            }
+            if (title == null || title.isEmpty()) {
+                throw new IllegalArgumentException("Title must not be null or empty");
+            }
+            if (!type.equals(type.toUpperCase())) {
```
robinst:
```suggestion
            if (!type.equals(type.toUpperCase(Locale.ROOT))) {
```

(Otherwise, e.g. in a Turkish locale `i` will uppercase to `İ`, which is probably unexpected.)

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
```diff
@@ -0,0 +1,117 @@
...
+                return Set.of();
+            }
+        });
+    }
+
+    /**
+     * Builder for configuring the alerts extension.
+     */
+    public static class Builder {
+        private final Map<String, String> customTypes = new LinkedHashMap<>();
```
robinst: Same here, no need for a `LinkedHashMap`

robinst (commented):
More feedback. Getting closer, but I think we can still simplify.

Neat idea to generate the test data by using the gh CLI.

robinst (approved):
Just one thing left that I missed before.

robinst:
Released in 0.28.0 now, thanks for the contribution 🎉!: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0280---2026-03-31

PR #422, opened 2026-03-30
add node depth limit

spirit-at-canva:
### Why

I want to limit the max depth of node traversal.

### What

We apply the limit when constructing the AST, by dropping deeper node generation (we collapse it into plain text).


robinst (commented):
Changes to `DocumentParser`, `Parser`, `ParserTest` LGTM, with some minor comments to address.

The visitor changes: Can you take them out of this PR? I'm not sure I want to include this by default on every visitor, and it's something that's pretty straightforward to add in subclasses (overriding the protected `visitChildren` method).

robinst:
Released in 0.28.0 now, thanks for the contribution 🎉!: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0280---2026-03-31

PR #425, opened 2026-04-01
Clean ups

miasma:
I've analyzed the code with IntelliJ's inspections and applied a couple of common refactorings.

 - redundant overrides add bloat and don't have any functional effect
 - redundant variable is used as return value and can be inlined
 - redundant label does not improve readability when there is only one label 
 - nested classes with static operate more efficiently
 - method references are potentially more efficient and more readable than anonymous classes
 - refactored API calls document the intent better (and can be used since the pom.xml defaults to jdk11)
 - redundant imports slow down build process

### DIFF
--- a/commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
+++ b/commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
@@ -6,8 +6,6 @@ import org.commonmark.ext.gfm.alerts.internal.AlertHtmlNodeRenderer;
 import org.commonmark.ext.gfm.alerts.internal.AlertMarkdownNodeRenderer;
 import org.commonmark.parser.Parser;
 import org.commonmark.renderer.NodeRenderer;
-import org.commonmark.renderer.html.HtmlNodeRendererContext;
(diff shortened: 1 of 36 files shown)


robinst:
Hey! Thanks for raising this. It currently doesn't pass the [Android compat build](https://github.com/commonmark/commonmark-java/actions/runs/23857230478/job/70735354680?pr=425):

```
/home/runner/work/commonmark-java/commonmark-java/commonmark/src/main/java/org/commonmark/internal/LinkReferenceDefinitionParser.java:106: error: cannot find symbol
> Task :app:compileDebugJavaWithJavac
        var removedSpans = List.copyOf(sourceSpans.subList(Math.max(sourceSpans.size() - lines, 0), sourceSpans.size()));
                               ^
  symbol:   method copyOf(List<SourceSpan>)
  location: interface List
```

Can you undo that particular change?

robinst:
@miasma ^?

miasma:
@robinst Hi, sorry, was a bit busy for the last 2 weeks.  I updated the branch and deleted that problematic commit. I wasn't sure which API level of Android is supported so it's maybe better to leave out those API changes for now.

miasma:
I don't think I have any additions to this anymore.

robinst:
Thanks! Merged.

Issue #429, opened 2026-04-29
Allow Adding Custom Titles to Alerts


rdestefa:
**Is your feature request related to a problem? Please describe.**

GFM alerts were added as part of https://github.com/commonmark/commonmark-java/issues/327. Some other Markdown renderers (notably [GitLab-flavored Markdown](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/)) extend this concept to allow specifying custom titles on the same line as the alert marker. For example:

```markdown
> [!TIP] Custom title
> Body text
```

**Describe the solution you'd like**

A toggle in `AlertsExtension` such as `allowCustomTitles` that would provide this behavior.

**Describe alternatives you've considered**

No real alternatives available currently.

rdestefa:
Made an initial attempt at implementing this here: #430

PR #430, opened 2026-04-29
Add Support for Custom Alert Titles

rdestefa:
Implements #429

Made an initial attempt at implementing support for custom titles in alerts (#429). I kept running into issues when trying to enhance the post processor because there's no access to an inline parser and it was susceptible to edge cases where inline nodes would overlap between the title and the body. For example:

```markdown
> [!TIP] 2*2 = 4
> But 3*3 = 9
```

The `*`s would create an `Emphasis` node which made splitting out the title text difficult, when in reality those lines should be parsed separately. I decided to try and make a custom `AlertBlockParser` instead to have more control.

I'm not sure if this is the right approach, but it seemed to work well for me. A few things I noticed, though, with this new parser:

- Alert markers take precedence over link reference definitions. This is opposite of what GitHub does, but in line with [what GitLab does](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%3E%20%5B!NOTE%5D%0A%3E%20Body%20text%0A%0A%5B!NOTE%5D%3A%20https%3A%2F%2Fexample.com%0A%5B!TIP%5D%3A%20https%3A%2F%2Fexample.com%0A%0A%3E%20%5B!TIP%5D%0A%3E%20Body%20text). My guess is that this isn't intentional by GitHub, and just a side effect of how they post-process after initial parsing has completed. To me this seems like an extremely rare edge case, and it's more likely the author wanted an alert than a link if they put a marker in that specific spot.
- Currently lazy continuation is not allowed between the title and the body text, for example:
  ```markdown
  > [!NOTE] Title
  Lazy body
  ```
  GitHub allows this, but [GitLab doesn't](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%3E%20%5B!NOTE%5D%20Title%0ALazy%20body). To me the latter seems correct because they are technically two separate nodes. This also seems to be in line with [this part of the CommonMark spec](https://github.github.com/gfm/#example-212:~:text=Laziness%20only%20applies%20to%20lines%20that%20would%20have%20been%20continuations%20of%20paragraphs%20had%20they%20been%20prepended%20with%20block%20quote%20markers.%20For%20example%2C%20the%20%3E%20cannot%20be%20omitted%20in%20the%20second%20line%20of). GitHub's support for this also seems unintentional and just a side-effect of it originally being parsed as `BlockQuote > Paragraph`, and the paragraph allows lazy continuation.
- This new parser has a configuration to allow alerts to be nested within other blocks (even other alerts). GitHub doesn't allow this, but [GitLab does](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%3E%20%5B!CAUTION%5D%20Custom%20title%201%0A%3E%20first%20layer%0A%3E%0A%3E%20%3E%20%5B!NOTE%5D%20Custom%20title%202%0A%3E%20%3E%20second%20layer%0A%3E%20%3E%0A%3E%20%3E%20%3E%20%5B!TIP%5D%20Custom%20title%203%0A%3E%20%3E%20%3E%20third%20layer). I could also separate this out into a different PR.
- Should alerts without a body be allowed? For example:
  ```markdown
  > [!NOTE]

  > [!NOTE] Title
  ```
  Currently it's not allowed. GitHub disallows this, but [GitLab allows it](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%3E%20%5B!CAUTION%5D%0A%0A%3E%20%5B!NOTE%5D%20Title).


rdestefa:
@robinst If you have a chance would you be able to take a look at this PR? Thanks!

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertBlockParser.java
```diff
@@ -0,0 +1,243 @@
...
+
+        /*
+         * Inline-parse the title in its own scope so delimiters are isolated
+         * from the body text. For example:
+         *
+         * > [!NOTE] 2*2 = 4
+         * > But 3*3 = 9
+         */
+        var titleNode = new AlertTitle();
+        inlineParser.parse(SourceLines.of(SourceLine.of(titleContent, null)), titleNode);
```
robinst: The `null` here means we're losing source positions of the inline content. To fix it, you need to change parsing of `titleContent` to not just return a `String`, but preserve the original `SourceLine` in `tryStartFresh`.

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertBlockParser.java
```diff
@@ -0,0 +1,243 @@
...
+        return BlockContinue.atColumn(newColumn);
+    }
+
+    @Override
+    public void parseInlines(InlineParser inlineParser) {
+        // Determine if there is any non-title body content.
+        if (block.getFirstChild() == null) {
+            /*
+             * Replace the Alert with a BlockQuote whose only paragraph contains
+             * the original first line text.
```
robinst: You're saying GitLab allows this case and keeps it as an alert? I think we can just do the same, then we don't need the demotion logic (which also loses source position information).

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertBlockParser.java
```diff
@@ -0,0 +1,243 @@
...
+                 */
+                afterGt = state.getIndex();
+            }
+
+            Matcher matcher;
+            if (customTitlesAllowed) {
+                matcher = ALERT_PATTERN_CUSTOM_TITLE.matcher(line.subSequence(afterGt, line.length()));
+            } else {
+                matcher = ALERT_PATTERN_NO_CUSTOM_TITLE.matcher(line.subSequence(afterGt, line.length()));
+            }
```
robinst: Do this instead?:

```suggestion
            var pattern = customTitlesAllowed ? ALERT_PATTERN_CUSTOM_TITLE : ALERT_PATTERN_NO_CUSTOM_TITLE;
            var matcher = pattern.matcher(line.subSequence(afterGt, line.length()));
```

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertHtmlNodeRenderer.java
```diff
@@ -35,7 +38,27 @@ protected void renderAlert(Alert alert) {
...
+             * > [!TIP] <!-- This won't appear --> <!---> <!-- Neither will this -->
+             * > Body text
+             *
+             * Then the reader will see a blank title. In this case, it's better
+             * to render the default title (but keep the comments for accuracy).
+             *
+             * These comments will be visible when rendered to other formats like
+             * Markdown, so this is only relevant for the HTML renderer.
+             */
+            if (isCommentOnlyTitle(first)) {
```
robinst: Does GitLab do this? Seems like a strange edge case to support.

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
```diff
@@ -108,6 +126,47 @@ public Builder addCustomType(String type, String title) {
...
+        public Builder allowCustomTitles() {
+            customTitlesAllowed = true;
+            return this;
+        }
+
+        /**
+         * Disallows custom titles on alerts. See {@link AlertTitle} for more information.
+         * @return {@code this}
+         */
+        public Builder disallowCustomTitles() {
```
robinst: Instead of two methods here and below, I'd prefer a single one with a boolean argument, so just `allowCustomTitles(true)` or `allowCustomTitles(false)`.

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertBlockParser.java
```diff
@@ -0,0 +1,243 @@
...
+
+            if (!matcher.matches()) {
+                return BlockStart.none();
+            }
+
+            String typeOriginalCase = matcher.group(1);
```
robinst: Style: In this and other places, can you use `var` where possible?

robinst (changes requested):
Nice work so far (and thanks for your patience)! Requested some changes, mainly around edge cases and preserving source position information.

boyshell:
您好！邮件已收到。

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertHtmlNodeRenderer.java
rdestefa: GitLab doesn't do this because they treat the entire title content as raw text instead of parsing inlines ([example](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%3E%20%5B!CAUTION%5D%20_No%20italics_%0A%0A%3E%20%5B!NOTE%5D%20**No%20bold**%0A%0A%3E%20%5B!IMPORTANT%5D%20%3C!--%20comment%20--%3E)). It seems like supporting inlines makes sense, however, so I did that. When I was playing around with it I noticed this edge case where the title would render but appear empty since it's just an HTML comment. I could see someone writing something like:

```markdown
> [!WARNING] <!-- Remove once this feature is GA -->
> This feature is experimental. Use at your own risk. 
```

If you don't think it's worth supporting though happy to remove.

robinst: Ok, so I was looking at GitLab's implementation (which seems to be comrak), and yes, it explicitly doesn't support formatting:

https://github.com/kivikakk/comrak/blob/main/src/tests/fixtures/alerts.md

So I went around looking for other implementations, and saw that Obsidian has [callouts](https://obsidian.md/help/callouts). Playing around with it, they **do** support formatting in the title.

So I think supporting inline formatting in the title is fine (but without the HTML comments special case). Can you also explicitly call it out (hah) in the Javadocs, that formatting is supported?

rdestefa: It's mentioned in the Javadocs for [`AlertTitle`](https://github.com/commonmark/commonmark-java/pull/430/changes#diff-7c35832956b50a1034bbbc3d225c5f086dc4065c9806fd43def42ff960d661f0R10-R19) and in the [alerts README](https://github.com/commonmark/commonmark-java/pull/430/changes#diff-3ee1554cbecaaa081e56c3ad59475b4ea05e812c8894dc36549eab0fca9a0321R52-R53). Is there another place you'd like me to mention it?

robinst: In `AlertsExtension` please, on the class and on the option.

rdestefa: Done!

robinst (approved):

Issue #386, opened 2025-06-30
Unable to override default delimiter processors


MerchantCalico:
**Is your feature request related to a problem? Please describe.**
I wish to override the default delimiter processors the sake of better supporting Discord's markdown quirks. However, I am left unable to override the default options without an exception being thrown.

**Describe the solution you'd like**
I would like to be able to override the two default delimiter processors with my own implementations of the two somehow within the parser builder.

**Describe alternatives you've considered**
The current solution for this is to copy the InlineParserImpl class and make your modifications to it, which is definitely not ideal.

Charchad:
I also noticed the same issue, in the `InlineParserImpl` class the **AsteriskDelimiterProcessor** and **UnderscoreDelimiterProcessor** are added by default in `calculateDelimiterProcessors`.

In my case I need to change the default behaviour, so text surrounded by * is bold, and by _ is italic, but as far as I can see this is not possible without redefining the whole inline parser. I can't add a new delimiter processor because I get an exception: "Cannot add two delimiter processors for char '_' and minimum length 1"

rdestefa:
I made a very rough attempt at this: https://github.com/commonmark/commonmark-java/pull/432

PR #432, opened 2026-05-02
Attempt at Supporting Overriding Core Delimiter Processors

rdestefa:
Quick attempt at supporting #386. Not sure if I've covered all edge cases or if there's a simpler way to do this, so might come back to this later.


robinst:
Instead of a separate configuration, can you explore how it would look if the normal `customDelimiterProcessor` would just be able to override built-in ones? I think we'd still want to error if a user accidentally specifies two processors with the same configuration, but not if they override a built-in one (i.e. add the custom ones first, then add the built-in ones if they don't have conflicts).

robinst (changes requested):
See comment above

Issue #391, opened 2025-08-07
Better YAML metadata parsing


zyxist:
**Is your feature request related to a problem? Please describe.**

I'm building a project, where the important part is parsing the YAML front matter in Markdown documents and map it to known DTO-s. I experienced multiple issues with the existing front matter extension for commonmark:

 * it supports only a small subset of YAML, unsuitable for my needs,
 * it produces the data structure that breaks Jackson when trying to map it to a DTO. The main issue is that `key: value` pair is mapped to a list with a single value. It also does not support other data types.

**Describe the solution you'd like**

I think that it would work much better, if YAML parsing were delegated to an external library. The extension could simply allow to "bring your own parser" via an interface to implement. In case of Jackson, this translates to calling a single method on `ObjectMapper`:

Pseudo-code:

```java
public class JacksonWrapper implements YamlParserProvider {
   private final ObjectMapper objectMapper;
   
   public JacksonWrapper(ObjectMapper objectMapper) {
      this.objectMapper = objectMapper;
   }

   public Map<String, Object> parseYaml(String input) {
      return objectMapper.readValue(input, Map.class);
   }
}

// ...
var parser = Parser.builder()
   .extensions(List.of(YamlFrontMatterExtension(JacksonWrapper(objectMapper))))
   .build();
```

In this way, commonmark does not have to have a dependency on any particular YAML parser.

**Describe alternatives you've considered**

Currently, I wrote a simple custom extension on my own. The thing that I don't have yet is better error handling, but the approach seems to work. But I would be happy, if such an extension were a part of Commonmark (less boilerplate code to maintain).

I briefly considered Commonmark to extend the parser in the existing extension, but I quickly realized that this is a complex task.

zyxist:
That would also work - the only challenge I see would be literal blocks that may contain `---` sequence as the first symbols in the line. You would be able to run the YAML parser only after processing of the entire Markdown document, so either you accept this trade-off or implement some minimal support just for literal blocks in YAML Front Matter extension.

But for me this is a corner case, and I could probably live with it.

zyxist:
Hi! I found some time to get back to the topic, and even prepared a contribution based on your idea to extract the string :).

Regarding your question, I was considering the following YAML example:

```yaml
key: value
multiline: '
---
string
'
more: keys
```

If we put this document into YAML front matter, the front matter parser terminates too early. However, I did extra research:

[...]

PR #434, opened 2026-05-11
Add support for Markdown rendering for YAML front matter

stupar73:
`commonmark-ext-yaml-front-matter` does not currently support rendering to Markdown, this PR adds that functionality.

The below block will be rendered at the top of the markdown file (containing the actual key-value pairs):
```yaml
---
# key: value pairs here
---

# Markdown content starts here
```

### DIFF
--- a/commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/internal/YamlFrontMatterMarkdownNodeRenderer.java
+++ b/commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/internal/YamlFrontMatterMarkdownNodeRenderer.java
@@ -1,0 +1,116 @@
+package org.commonmark.ext.front.matter.internal;
+
+import java.util.List;
+import org.commonmark.ext.front.matter.YamlFrontMatterNode;
+import org.commonmark.node.Node;
+import org.commonmark.renderer.markdown.MarkdownNodeRendererContext;
+import org.commonmark.renderer.markdown.MarkdownWriter;
+
+public class YamlFrontMatterMarkdownNodeRenderer extends YamlFrontMatterNodeRenderer {
+
+    private final MarkdownWriter writer;
+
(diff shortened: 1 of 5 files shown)


commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/internal/YamlFrontMatterMarkdownNodeRenderer.java
```diff
@@ -0,0 +1,116 @@
...
+                || value.contains("\"")
+                // Unbalanced flow-style list
+                || value.startsWith("[")
+                // Unbalanced flow-style mapping
+                || value.startsWith("{");
+    }
+
+    private boolean isFlowCollection(String value) {
+        return (value.startsWith("[") && value.endsWith("]"))
+                || (value.startsWith("{") && value.endsWith("}"));
```
robinst, line 114: All this special handling is a bit unfortunate, we should probably never have started trying to parse the YAML ourselves (but treat it as an opaque blob instead). See also this issue:
- https://github.com/commonmark/commonmark-java/issues/391

Having said that, it is what we currently do so what you have here LGTM.

robinst (commented):
LGTM, thank you!

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/internal/YamlFrontMatterMarkdownNodeRenderer.java
stupar73: Yeah agreed, it feels a bit like it shouldn't be this libraries responsibility to handle the content of the YAML front matter.

But as you say, that's what we currently do, and I think this is an net positive from the current state.

robinst:
Released in 0.29.0 now: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0290---2026-06-20

Issue #431, opened 2026-05-02
Allow Removing Certain GFM Alert Types


rdestefa:
**Is your feature request related to a problem? Please describe.**

Sometimes a parser/renderer will only want to enable a subset of GFM alert types or have a completely custom set of allowed types. Currently there's no way to disallow any of the standard GFM types.

**Describe the solution you'd like**

A helper function on the builder like `removeType` would be useful here.

rdestefa:
Made an attempt at this here: https://github.com/commonmark/commonmark-java/pull/435

robinst:
Merged the PR. Want me to cut a release, or do you have any other related changes?

rdestefa:
Thanks for merging! I don't have any other related changes, so feel free to cut a release.

rdestefa:
Ended up noticing one more missing feature for the alerts extension: #437

robinst:
Release 0.29.0 now with all the improvements: https://github.com/commonmark/commonmark-java/blob/main/CHANGELOG.md#0290---2026-06-20

PR #435, opened 2026-06-03
Allow Removing Certain GFM Alert Types

rdestefa:
Implements #431

### DIFF
--- a/commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
+++ b/commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
@@ -155,30 +155,6 @@ public class AlertsExtension implements Parser.ParserExtension, HtmlRenderer.Htm
         }
 
         /**
-         * Adds a custom alert type with a default title.
-         * <p>
-         * This can also be used to override the default title of standard GFM types
-         * (e.g., for localization).
-         *
-         * @param type the alert type (must be uppercase)
-         * @param title the default title for this alert type
-         * @return {@code this}
-         */
-        public Builder addCustomType(String type, String title) {
-            if (type == null || type.isEmpty()) {
-                throw new IllegalArgumentException("Type must not be null or empty");
-            }
-            if (title == null || title.isEmpty()) {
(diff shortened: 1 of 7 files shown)


rdestefa:
@robinst Another quick PR if you have time. No rush though!

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/AlertsExtension.java
```diff
@@ -125,7 +138,34 @@ public Builder addCustomType(String type, String title) {
...
+        }
+
+        /**
+         * Removes alert types from the allowed list.
+         *
+         * @param types the alert types to remove (must be uppercase)
+         * @return {@code this}
+         * @see AlertsExtension#STANDARD_TYPES
+         */
+        public Builder removeTypes(String... types) {
```
robinst: I'm wondering if it would be nicer if the API was just `setAllowedTypes(Map<String, String>)` instead, which would allow you to set the exact types you want (including custom ones). It would just override the whole map. With the current API in some cases you'd have to do remove+add. What do you think?

rdestefa: Thanks for the suggestion! Agree that `setAllowedTypes(Map<String, String>)` would be better than `removeTypes(String...)`. I think `addCustomType` would still be good to keep so users who just want to add 1 or 2 more types don't need to completely recreate the set, but I made two commits so you can see which you think works best:

- https://github.com/commonmark/commonmark-java/pull/435/changes/d6657350ffaa8122894e1e9ae92bb352d08543ce - Replaces `removeTypes` with `setAllowedTypes`, but keeps `addCustomType`
- https://github.com/commonmark/commonmark-java/pull/435/changes/9c813f6053400b06ce3dcb267c2b4111f3ea8e34 - Additionally removes `addCustomType`

robinst: Ah yeah I think `addCustomType` can stay.

rdestefa: Done!

Issue #436, opened 2026-06-10
Add Text Content Rendering Support for Alerts


rdestefa:
The `AlertsExtension` currently only supports HTML and Markdown rendering. It would be great to have text rendering support as well.

rdestefa:
Made a PR for this here: #437

PR #437, opened 2026-06-10
Add Text Content Rendering Support for Alerts

rdestefa:
Implements #436

### DIFF
--- a/commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertHtmlNodeRenderer.java
+++ b/commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertHtmlNodeRenderer.java
@@ -24,7 +24,8 @@ public class AlertHtmlNodeRenderer extends AlertNodeRenderer {
     @Override
     protected void renderAlert(Alert alert) {
         var type = alert.getType();
-        if (!allowedTypes.containsKey(type)) {
+        var defaultTitle = allowedTypes.get(type);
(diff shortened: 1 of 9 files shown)


commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertHtmlNodeRenderer.java
```diff
@@ -24,6 +24,9 @@ public AlertHtmlNodeRenderer(HtmlNodeRendererContext context, Map<String, String
     @Override
     protected void renderAlert(Alert alert) {
         var type = alert.getType();
+        if (!allowedTypes.containsKey(type)) {
+            throw new IllegalStateException("Unknown alert type: " + type);
+        }
```
robinst: Instead of a separate `containsKey` and `get`, can you write these like:


```suggestion
        var typeTitle = allowedTypes.get(type);
        if (typeTitle == null) {
            throw new IllegalStateException("Unknown alert type: " + type);
        }
        
        // later use typeTitle
```

robinst (commented):
Just one comment, otherwise LGTM

commonmark-ext-gfm-alerts/src/main/java/org/commonmark/ext/gfm/alerts/internal/AlertHtmlNodeRenderer.java
rdestefa: Done!

PR #445, opened 2026-08-01
Better YAML front matter parsing in commonmark-ext-yaml-front-matter

zyxist:
Last year, I reported the issue #391 - I found some time to turn it into the contribution, based on the outcome of the discussion. I will be happy to discuss the approach and apply any improvements, if necessary.

### Summary

The PR splits the YAML front matter processing into two parts:

 1. Finding front matter delimiters
 2. Processing the captured front matter content with the new concept of `YamlFrontMatterExtractor`.

I moved the existing built-in parser of YAML subset into one of the two extractors: `YamlDataExtractor`. The other extractor is `YamlContentExtractor` that extracts the front matter as a string for further processing with other tools.

### Compatibility

The changes are backward-compatible. `YamlDataExtractor` is the default extractor, and one must explicitly initialize the extension with `YamlContentExtractor` to get the string content. The visitor and Markdown renderer support both modes.

### Naming

For the existing parser, I strived to keep consistent naming to help understand the connection: `YamlFrontMatterVisitor.getData()` (existing method) => `YamlDataExtractor`.


README.md
```diff
@@ -407,7 +408,36 @@ literal: |
...
+List<Extension> extensions = List.of(YamlFrontMatterExtension.create());
+Parser parser = Parser.builder()
+    .extensions(extensions)
+    .build();
+
+Node document = parser.parse(markdownDocument);
+Map<String, List<String>> frontMatter = YamlFrontMatterVisitor.readData(document);
+```
+
+Alternatively, you can use initialize the extension with `YamlContentExtractor` that saves the YAML front matter content
```
robinst: typo "use initialize"

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterContent.java
```diff
@@ -0,0 +1,19 @@
+package org.commonmark.ext.front.matter;
+
+import org.commonmark.node.CustomNode;
+
+public class YamlFrontMatterContent extends CustomNode {
```
robinst: So the existing one is called `YamlFrontMatterNode`. I'm not sure if just "Content" is enough to distinguish the two. What do you think about naming this one `YamlFrontMatterRawContent` (and the method `getLiteral()` (like what code blocks call their content)? Also, can you add some Javadoc here and for `YamlFrontMatterNode`?

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterExtractor.java
```diff
@@ -0,0 +1,13 @@
+package org.commonmark.ext.front.matter;
+
+import org.commonmark.parser.block.BlockContinue;
+
+public interface YamlFrontMatterExtractor {
```
robinst: Hmm not sure about calling this an "extractor", it feels like what the visitor does sounds more like extracting. We already have a good word for turning some input into nodes, which is a parser. Should we just call this `FrontMatterParser`?

For the name of the implementations I have similar concerns (them not being specific enough). With the above change, they would be:

- `BasicYamlParser`
- `RawContentParser`

(Note that the raw content parser could also be used for e.g. TOML or other syntax of front matter.)

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterExtractor.java
```diff
@@ -0,0 +1,13 @@
...
+    void onNextLine(YamlFrontMatterBlock block, CharSequence line);
+
+    BlockContinue onBlockEnd(YamlFrontMatterBlock block);
```
robinst: I would remove the return type here, as anything else than `BlockContinue.finished()` doesn't make sense.

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterExtractor.java
```diff
@@ -0,0 +1,13 @@
+package org.commonmark.ext.front.matter;
+
+import org.commonmark.parser.block.BlockContinue;
+
+public interface YamlFrontMatterExtractor {
+    void onNextLine(YamlFrontMatterBlock block, CharSequence line);
```
robinst: Can you change the line to `SourceLine` instead (which is what `parserState.getLine()` returns)?

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterVisitor.java
```diff
@@ -3,28 +3,89 @@
...
+    public static Map<String, List<String>> readData(Node document) {
+        YamlFrontMatterVisitor visitor = new YamlFrontMatterVisitor();
+        document.accept(visitor);
+        return visitor.getData();
+    }
+
+    /**
+     * Reads the YAML Front Matter metadata as a string, if the Markdown
+     * document has the YAML Front Matter and the extension uses
+     * {@link YamlContentExtractor} (default).
```
robinst:
```suggestion
     * {@link YamlContentExtractor}.
```

robinst (changes requested):
Thanks for working on this! I have a few things I'd like to see changed but the direction is good.

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterExtractor.java
zyxist: You're right, I struggled to find the right naming here. I initially rejected "parser", because I had a very similar name `YamlFrontMatterBlockParser` next to it, but the term is perfect here. I will take a look :)

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterExtractor.java
zyxist: Actually, it allows the parser to handle the following use case described here, if someone needs it:

https://github.com/commonmark/commonmark-java/issues/391#issuecomment-5151487352

But this is a minor issue with easy workaround, so I can either simplify the interface or document the purpose in the javadoc.

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/YamlFrontMatterExtension.java
```diff
@@ -18,21 +22,38 @@
...
+ * {@link YamlFrontMatterVisitor#readRawContent(Node)} to process it with other tools.
+ *
+ * <p>Implement {@link FrontMatterParser} interface and the corresponding factory to
+ * parse the front matter with a custom parser.
  */
 public class YamlFrontMatterExtension
         implements Parser.ParserExtension, MarkdownRenderer.MarkdownRendererExtension {
 
-    private YamlFrontMatterExtension() {}
+    private final FrontMatterParser.Factory yamlExtractorFactory;
```
robinst: Looks like you missed a rename here (and corresponding parameters).

commonmark-ext-yaml-front-matter/src/main/java/org/commonmark/ext/front/matter/parser/FrontMatterParser.java
robinst, line 1: Should this also be in the `parser` subpackage?

PR #448, opened 2026-08-05
fix #447

### LINKED ISSUE #447: Avoid quadratic-time parsing for inline HTML
### Summary

Every `<` in a document makes the inline parser look for the end of the construct it might start, and the scan is thrown away when that end is not there. On input made of `<` whose construct never ends, each one walks to the end of the input, so parsing is quadratic in the size of the input. Roughly 800 KB of `x <!-- >` keeps a core busy for over a minute, so a single request can stall a worker in any application that renders untrusted Markdown (comments, chat, wikis).

### Details

[`HtmlInlineParser`](https://github.com/commonmark/commonmark-java/blob/7a68e9204c8f4e521f8206aff3b64361e1a605b1/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java) is triggered by `<` and has four scans of this shape:

| Construct | Looks for | Code |
|---|---|---|
| processing instruction | `?>` | [L153](https://github.com/commonmark/commonmark-java/blob/7a68e9204c8f4e521f8206aff3b64361e1a605b1/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java#L153) |
| comment | `-->` | [L177](https://github.com/commonmark/commonmark-java/blob/7a68e9204c8f4e521f8206aff3b64361e1a605b1/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java#L177) |
| CDATA section | `]]>` | [L196](https://github.com/commonmark/commonmark-java/blob/7a68e9204c8f4e521f8206aff3b64361e1a605b1/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java#L196) |
| declaration | `>` | [L215](https://github.com/commonmark/commonmark-java/blob/7a68e9204c8f4e521f8206aff3b64361e1a605b1/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java#L215) |

Each of them ends up in [`Scanner.find`](https://github.com/commonmark/commonmark-java/blob/7a68e9204c8f4e521f8206aff3b64361e1a605b1/commonmark/src/main/java/org/commonmark/parser/beta/Scanner.java#L196), which advances one character at a time and only gives up at the end of the input:

```java
while (scanner.find('-') >= 0) {   // no `-->` ahead, so this walks to the end of the input
    if (scanner.next("-->")) {
        return true;
    } else {
        scanner.next();
    }
}
```

The candidate is then rejected, the position is reset, and the next `<` starts the same scan over again. With n openers that is n scans of average length n/2. Plain paragraph text is enough to reach it:

| Input, repeated | Size at 100,000 | 25,000 | 50,000 | 100,000 | 100,000 with the patch |
|---|---|--:|--:|--:|--:|
| `x <? >` | 586 KB | 1.7 s | 6.6 s | 26.7 s | 0.08 s |
| `x <!-- >` | 781 KB | 4.6 s | 19.2 s | 76.7 s | 0.07 s |
| `x <CDATA[ >` | 1270 KB | 3.5 s | 15.3 s | 63.0 s | 0.13 s |

Measured with `Parser.builder().build()` and `HtmlRenderer.builder().build()` on JDK 25, current `main` (`7a68e920`). Times roughly quadruple when the input doubles.

Each payload deliberately contains a `>` close to every `<`, so that the separately reported autolink scan for `>` stays cheap and only the inline HTML scanning is measured. It also starts with text so the line doesn't begin with `<`, which would parse as an HTML block instead.

A PR will shortly follow.

### Credits

Found by [Team Atlanta.  
Collected and verified by [OSTIF](https://ostif.org) using AI with their own harness.  
Manually verified, fixed, and reported by [Shielder](https://www.shielder.com).

fromVeeko:
As requested, here the PR with a fix proposal.

### DIFF
--- a/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java
+++ b/commonmark/src/main/java/org/commonmark/internal/inline/HtmlInlineParser.java
@@ -42,6 +42,16 @@ public class HtmlInlineParser implements InlineContentParser {
                     .c('`')
                     .build();
 
+    // A scan that ran to the end of the input without finding the terminator it was looking for
+    // proves that no later scan in the same inline snippet can find it either (inline parsing only
+    // moves forward). Remembering that stops each `<` from scanning the rest of the input again,
+    // which would be quadratic for input like `<!--` repeated many times. A parser is created for
+    // each inline snippet, so these don't need to be reset.
+    private boolean noProcessingInstructionEnd;
+    private boolean noCommentEnd;
+    private boolean noCdataEnd;
+    private boolean noDeclarationEnd;
+
     @Override
     public ParsedInline tryParse(InlineParserState inlineParserState) {
         Scanner scanner = inlineParserState.scanner();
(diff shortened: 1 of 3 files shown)


robinst:
Great, thank you!

PR #456, opened 2026-09-03
MarkdownWriter: append unescaped spans in bulk

tballison:
During profiling after the release of Tika 4.0.0, we found we could improve write speed quite a bit in cases where there weren't a lot of characters that needed to be escaped.

:robot: 
```
text() always passes a CharMatcher, so every text literal was written 
to the Appendable one character at a time. Scan for the next character 
that needs escaping and append the clean span in one call instead.
```
## JMH, whole CommonMark spec (`SpecBenchmark`, JDK 17, `-f 3 -wi 8 -i 12`)

| Benchmark | before | after | |
|---|---|---|---|
| `renderMarkdownWholeSpec` (to `StringBuilder`) | 836.5 ± 34.3 ops/s | 956.7 ± 11.3 ops/s | +14% |
| `renderMarkdownWholeSpecToWriter` | 401.9 ± 2.5 ops/s | 837.2 ± 20.1 ops/s | **2.08x** |

### Rendering 4.7 MB of plain prose (best-of-22 after warmup, JDK 17)

| Target `Appendable` | before | after | |
|---|---|---|---|
| `OutputStreamWriter` | 107 ms | 14.5 ms | **7.4x** |
| `StringWriter` | 75 ms | 12.3 ms | **6.1x** |
| `StringBuilder` | 26 ms | 12.7 ms | **2.0x** |

### Same measurement, 4.0 MB of the spec (escape-dense, closer to a worst case)

| Target `Appendable` | before | after | |
|---|---|---|---|
| `OutputStreamWriter` | 54.9 ms | 25.8 ms | 2.1x |
| `StringWriter` | 46.1 ms | 21.5 ms | 2.1x |
| `StringBuilder` | 24.2 ms | 20.8 ms | 1.2x |


commonmark/src/main/java/org/commonmark/renderer/markdown/MarkdownWriter.java
```diff
@@ -158,8 +158,23 @@ private void write(String s, CharMatcher escape) {
                 // Normal fast path
                 buffer.append(s);
             } else {
-                for (int i = 0; i < s.length(); i++) {
-                    append(s.charAt(i), escape);
+                // Append the characters that don't need escaping in bulk, a span at a time.
+                // Appending one character at a time is a lot slower for most Appendable
+                // implementations (e.g. Writer, where each char is a synchronized call).
```
roxspring: 💭 Worth extracting an `append(String, CharMatcher) throws IOException` method? Often seems that a comment like this is a signal that a method extraction is available, and the new loop+if looks very self-contained.

